PR #2, opened 2014-08-09
add samples

scorpiovn:


kuujo:
Sample code will have to be updated for `0.6` which will be officially released very soon.

PR #62, opened 2015-11-22
Support recovery of resource instance sessions

kuujo:
This is the final significant addition to Atomix before the 1.0 release. This PR implements a custom `RecoveryStrategy` that recovers the client's session and all open resource instances. This requires that when the client's session is expired, the client is opened and then each individual resource instance is opened. In order to prevent cases where a resource is deleted while the client is disconnected, we only reopen instances that still exist.


madjam:
This is a very important feature. Thanks for implementing this.
Looks good to me.
:+1:

jhalterman:
Late, but big :+1: Now to get this into atomix-jepsen.

PR #63, opened 2015-11-25
Fix issue that main thread doesn't wait for completion of assertion

oshothebig:
I found the main thread doesn't wait for completion of the assertion in DistributedAtomicLongTest. It causes all tests always pass without verifying assertions.


jhalterman:
Good stuff - thanks for this!

PR #64, opened 2015-11-27
Ensure locks are released when sessions are closed

kuujo:
This PR addresses a *critical bug* in `LockState` wherein locks were never released in the event that a client's session was closed while it held a lock. The fix is to add a `close` method to `LockState` to release the lock of a closed session and grant the lock to the next commit in the queue.


AaronWo:
Hi, I'm using Atomix 1.0.8 and run into a situation:
My process got a lock _A_ and exited by accident. When I tried restarted the process and it tried to get lock _A_ again, it hung there.
Is this the same problem with yours?

Issue #59, opened 2015-11-02
Support for listing created resources


madjam:
This is useful from an administration stand point. The basic idea is to support an API in `io.atomix.manager.ResourceManager` for listing the resources that exist. Note that depending on the number of resources this operation may be expensive.

jhalterman:
I could see this feeding a future REST API, where we may want to support some simple filtering of listed resources by name or type.

kuujo:
Agree. Like you said, some thought has to go into this for cases where large numbers of resources exist (which is certainly possible since the overhead of a resource is just an entry in the log and an object in memory). Also relevant to this is just to provide more context about individual resources. There's currently no way to get the type of a resource, for instance. So, `exists("foo")` doesn't really tell me whether it's safe to try to use e.g. a `DistributedLock` for that resource, though I suppose attempting to create a `DistributedLock` with that name would fail if the state machines don't match. Alternatively, it would be possible to allow one instance of each resource for each key, but that could become confusing.

madjam:
Support added in #70

PR #70, opened 2015-12-02
Support for listing created resources

madjam:
(#59) Adds `Atomix#keys()` method to get a listing of resource keys. A variant of this method takes the resource type as input parameter and returns relevant resource keys.


kuujo:
Oooh nice love it.

:+1:

PR #71, opened 2015-12-02
Renamed atomix-atomic to atomix-primitives.

jhalterman:
This was necessary since atomix supports serializable consistency and therefore the word "atomic" was misleading.

Docs should require some updates as well, but we can catch them when we go over them (this week?)


madjam:
How about variables instead of primitives? Personally I tend to view primitives as "distributed programming primitives" and that includes everything under `collections`, `coordination` and the existing `atomic` package.

jhalterman:
That is a good point. Variables seems OK, but not quite perfect either. I'm not 100% on any name for this one...

jhalterman:
@kuujo What say ye? I was also thinking this package could be called `registers`, but register is yet another name with different contextual meanings.

kuujo:
Yeah, there's no good option. Primitives are sort of two different things. Primitives from a language's perspective (simple types) are different than primitives from a system's perspective (building blocks). I think variables may be a bit of an improvement. 

The other option is combining this module with collections. I think that may make some sense. For instance, we could potentially allow references to these "primitives" to be stored in the complex types, e.g. a set of counters. But that could still be done by e.g. adding a `Referenceable` interface to the `resource` module for identifying types that can be referenced, which is admittedly more interesting. 

Anyways, I guess variables or values or primitives would be the obvious options? I think the concept of "variables" or "values" makes even more sense again if resources can be referenced.

jhalterman:
Agree re: `Referenceable` and the resource module. If we were to combine primitives and collections though, collections wouldn't seem an appropriate name for `DistributedLong`. Let's assume resources can be referenced at some point, but for now - let's do variables. I think it's more definitive than values (plus in functional languages, a value is a constant whereas a variable is more like a register). Good?

kuujo:
Sure.

PR #75, opened 2015-12-06
Update Atomix usage of Copycat APIs

kuujo:
This PR updates usage of Copycat APIs including most importantly `CopycatServer` and `CopycatClient`. Recent API changes have removed `RaftClient` and `RaftServer` before the release.


jhalterman:
:+1:

PR #76, opened 2015-12-07
Log compaction updates

kuujo:
This PR updates all Atomix code to comply with the changes made in Copycat to log compaction processes. These changes include adding `CompactionMode`s to relevant commands, supporting `Snapshottable` state machines, and implementing `SessionListener` for state machines that need to listen for sessions being registered/unregistered/expired/closed.

### DIFF
--- a/collections/src/main/java/io/atomix/collections/state/MapCommands.java
+++ b/collections/src/main/java/io/atomix/collections/state/MapCommands.java
@@ -40,6 +40,10 @@ public class MapCommands {
    * Abstract map command.
    */
   public static abstract class MapCommand<V> implements Command<V>, CatalystSerializable {
+    @Override
+    public CompactionMode compaction() {
+      return CompactionMode.QUORUM_CLEAN;
+    }
 
     @Override
     public void writeObject(BufferOutput<?> buffer, Serializer serializer) {
@@ -224,8 +228,8 @@ public class MapCommands {
     }
 
     @Override
-    public PersistenceLevel persistence() {
-      return ttl > 0 ? PersistenceLevel.PERSISTENT : PersistenceLevel.EPHEMERAL;
+    public CompactionMode compaction() {
+      return ttl > 0 ? CompactionMode.FULL_SEQUENTIAL_CLEAN : CompactionMode.FULL_CLEAN;
     }
 
     /**
@@ -348,8 +352,8 @@ public class MapCommands {
     }
 
     @Override
-    public PersistenceLevel persistence() {
-      return PersistenceLevel.PERSISTENT;
+    public CompactionMode compaction() {
+      return CompactionMode.FULL_SEQUENTIAL_COMMIT;
     }
   }
 
@@ -367,8 +371,8 @@ public class MapCommands {
     }
 
     @Override
-    public PersistenceLevel persistence() {
-      return PersistenceLevel.PERSISTENT;
+    public CompactionMode compaction() {
+      return CompactionMode.FULL_SEQUENTIAL_COMMIT;
     }
   }
 
@@ -451,8 +455,8 @@ public class MapCommands {
   public static class Clear extends MapCommand<Void> {
 
     @Override
-    public PersistenceLevel persistence() {
-      return PersistenceLevel.PERSISTENT;
+    public CompactionMode compaction() {
+      return CompactionMode.FULL_SEQUENTIAL_COMMIT;
     }
   }
 
--- a/collections/src/main/java/io/atomix/collections/state/MapCommands.java
+++ b/collections/src/main/java/io/atomix/collections/state/MapCommands.java
@@ -229,7 +229,7 @@ public class MapCommands {
 
     @Override
     public CompactionMode compaction() {
-      return ttl > 0 ? CompactionMode.FULL_SEQUENTIAL_CLEAN : CompactionMode.FULL_CLEAN;
+      return ttl > 0 ? CompactionMode.FULL_SEQUENTIAL_CLEAN : CompactionMode.QUORUM_CLEAN;
     }
 
     /**
--- a/collections/src/main/java/io/atomix/collections/state/MultiMapCommands.java
+++ b/collections/src/main/java/io/atomix/collections/state/MultiMapCommands.java
@@ -42,6 +42,10 @@ public class MultiMapCommands {
    * Abstract map command.
    */
   public static abstract class MultiMapCommand<V> implements Command<V>, CatalystSerializable {
+    @Override
+    public CompactionMode compaction() {
+      return CompactionMode.QUORUM_CLEAN;
+    }
 
     @Override
     public void writeObject(BufferOutput<?> buffer, Serializer serializer) {
@@ -291,8 +295,8 @@ public class MultiMapCommands {
     }
 
     @Override
-    public PersistenceLevel persistence() {
-      return ttl > 0 ? PersistenceLevel.PERSISTENT : PersistenceLevel.EPHEMERAL;
+    public CompactionMode compaction() {
+      return ttl > 0 ? CompactionMode.FULL_SEQUENTIAL_CLEAN : CompactionMode.FULL_CLEAN;
     }
 
     /**
@@ -364,8 +368,8 @@ public class MultiMapCommands {
     }
 
     @Override
-    public PersistenceLevel persistence() {
-      return PersistenceLevel.PERSISTENT;
+    public CompactionMode compaction() {
+      return CompactionMode.FULL_SEQUENTIAL_COMMIT;
     }
   }
 
@@ -391,8 +395,8 @@ public class MultiMapCommands {
     }
 
     @Override
-    public PersistenceLevel persistence() {
-      return PersistenceLevel.PERSISTENT;
+    public CompactionMode compaction() {
+      return CompactionMode.FULL_SEQUENTIAL_COMMIT;
     }
 
     @Override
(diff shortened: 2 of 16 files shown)


madjam:
Changes look good.
:+1: 

I know you explained this before but I'm not able to look it up. Can you please explain the rationale behind the `FULL_SEQUENTIAL_CLEAN` and `FULL_SEQUENTIAL_COMMIT` CompactionModes? When is FULL_SEQUENTIAL_CLEAN/COMMIT appropriate vs FULL_CLEAN/COMMIT?

kuujo:
@madjam FYI you may have noticed, but I went a little rogue over the weekend. There have been a lot of huge pull requests and a lot of code added to Copycat over the last few days. In the interest of pushing forward on development, I broke my own rules and unilaterally merged a bunch of pull requests into both projects. These pull requests complete *all* of the features that will be present in the 1.0 release, after which rules certainly cannot be broken. But I think we can get much more done in small pull requests fixing minor issues with the changes than with trying to understand the major changes that went into support for snapshots and improvements to configuration changes before that. So, once this PR and the resource reference PR are merged, both projects will be in a feature freeze until the 1.0 release and all work will go into bug fixes and documentation. I'll push a release candidate once both are stabilized. But enough administrative talk.

I will document the compaction stuff likely some time today, but here's the gist of it...

In my view, the addition of `CompactionMode.SNAPSHOT` to Copycat allows for a little more complexity in the options for log cleaning in state machines because snapshots offer a much simpler implementation option. Previously, state machines only had essentially a non-tombstone and tombstone type command. Non-tombstone commands can be removed from the log once `clean`ed and tombstone commands can be removed once `clean`ed and stored on all servers. These two requirements are represented in `CompactionMode` as `QUORUM_COMMIT`/`QUORUM_CLEAN` and `FULL_COMMIT`/`FULL_CLEAN` respectively.

`QUORUM` means a command must be stored on a majority of servers before it can be compacted (the base Raft requirement) and `FULL` means a command must be stored on all servers before it can be compacted.

`COMMIT` means a command can be compacted once it's applied to the state machine on any given server and `CLEAN` means a command must be explicitly `clean`ed by the state machine to be removed from the log. For instance, a command that removes a key from a map could use the `COMMIT` suffix since it doesn't need to be retained in the state machine after the key is removed, but a command that sets a key would use `CLEAN` since the command contributes to the state machine's state until it's overwritten or removed. So, the four commands we have thus far are:
* `QUORUM_COMMIT` - stored on a majority of servers and applied to the state machine
* `QUORUM_CLEAN` - stored on a majority of servers and cleaned by the state machine
* `FULL_COMMIT` - stored on all servers and applied to the state machine
* `FULL_CLEAN` stored on all servers and cleaned by the state machine


The final two `CompactionMode`s indicate requirements for how the command should be removed from the log with respect to other commands in the log. Entries with any of the above four compaction modes can take place during either minor or major compaction. Remember, minor compaction can compact any subset of the segments in the log. So, a minor compaction process might compact segment 10 without compacting segments 1-9. But this poses a problem for removing tombstones from the log. In order for a tombstone to be removed from the log, we must be sure that all prior commands for which it removed state are removed from the log first. If the map key `foo` is set 10 times and then removed, the 10 `set` commands need to be removed before the `remove` command, but those commands may be spread out across many segments of the log. The two properties we do know are:
* The `remove` command appears later in the log than all state that it deletes
* All commands/state deleted by the `remove` command are `clean`ed before the `remove` command itself is cleaned
Thus, in order to ensure that all prior related commands are removed from the log before the `remove` command itself is removed, we clean the log sequentially during major compaction. The major compaction process iterates through all segments of the log up to the global compaction index to ensure that when a tombstone is encountered, all prior related commands will already have been removed from the log. The major compaction process takes an immutable snapshot of the state of cleaned entries in all segments to ensure they don't change during the compaction process (an earlier command is not `clean`ed after its segment is already compacted). So, the two `SEQUENTIAL` `CompactionMode`s essentially ensure that commands are removed only during major compaction.

There's also one other detail that is handled internally. The log compaction process will always only compact entries up to the highest index for which event messages have been received by all clients. This ensures that commands are *always* retained in the log so long as any session event messages resulting from the commands have not yet been acknowledged by clients. The snapshot is only stored once all events up to the snapshot index have been received, and so in the event of a failure and replay, the log will replay the actual commands that publish events to the session. But this is an internal implementation detail that is not represented in the `CompactionMode` options and is enforced for all commands.

In practice, what the `CompactionMode`s look like is this:
* Commands for which state is not held in the state machine can be marked with a `*_COMMIT` compaction mode
* Commands for which the state is held in the state machine for any amount of time must be marked with a `*_CLEAN` compaction mode and `clean`ed when no longer needed by the state machine
* Commands which should always be applied on all servers should be marked `FULL_*`. This could be useful for a variety of reasons
* Commands which overwrite prior state but are themselves immediately removed (tombstones) must be `FULL_SEQUENTIAL_*`

kuujo:
Here are the two `switch` statements that handle minor and major compaction of specific `CompactionMode`s:

https://github.com/atomix/copycat/blob/master/server/src/main/java/io/atomix/copycat/server/storage/compaction/MinorCompactionTask.java#L124
https://github.com/atomix/copycat/blob/master/server/src/main/java/io/atomix/copycat/server/storage/compaction/MajorCompactionTask.java#L215

jhalterman:
:+1:

madjam:
@kuujo Thank you for the detailed explanation.

I know we discussed this in a different thread and I was initially in favor of defining the options in terms of how the log gets compacted. But as I'm reading through the description I am a little worried if the minor differences between commit vs clean, full vs quorum and sequential vs non will be lost on some of the folks writing state machines. As a state machine writer the details that I know for sure are the semantics of the state machine itself such as a whether a command adds/removes/updates state or whether some operations are auto-expiring (ttl>0). If it possible for a state machine writer to just specify "how" a particular command mutates that state machine and then for Copycat to that take specification and map it to a particular compaction option then it might help prevent subtle bugs from creeping in. 

Perhaps I'm being a little naive in assuming all state machine types and their commands neatly fall into a a small set of categories and each of those categories map 1:1 to a compaction mode. But I felt it worth considering it first. If you already thought about it did you encounter any difficulties in classifying the commands?

kuujo:
I agree. I think it would be ideal to express compaction in this way internally - i.e. in `Compaction.Mode` - and express how a command behaves within the state machine externally. But I am also a little worried that that could limit state machines to a specific subset of behaviors.

Then again, I suppose multiple Command "types" could relate to a single type of compaction. For instance (I'm calling it "type" as a placeholder:
* `Command.Type.EXPIRING` could represent state that is eventually removed from the state machine (a TTL)
* `Command.Type.DELETE` could represent a command that deletes state
* `Command.Type.PERSISTENT` could represent a command that is immediately persisted by the state machine
* `Command.Type.MESSAGE` could represent a command that simply triggers a message to a session
* `Command.Type.UPDATE` could represent a command that updates state machine state

Respectively, those command types would point to `FULL_SEQUENTIAL_CLEAN`, `FULL_SEQUENTIAL_COMMIT`, `FULL_COMMIT`, `QUORUM_COMMIT`, and `QUORUM_CLEAN`. I actually *love* this idea. Ultimately, both could be exposed. That is, `Command.Type` could be an enum that provides a `CompactionMode`, and any additional use cases can be added even if they ultimately represent a compaction mode that's already covered. In that sense, I don't actually think we're limiting state machines to any specific behaviors. The `Command.Type` (or whatever) enum can just be an implementation of an interface that provides a `CompactionMode`, and so it can be extended or the `CompactionMode` can be explicitly overridden if necessary.

Thoughts?

kuujo:
@madjam the idea that this could limit potential use cases is exactly why I've avoided it, but presumably user feedback can simply promote coverage of more use cases, and it obviously makes for an API that's an order of magnitude simpler.

kuujo:
I'll open a PR for some form of this. I actually think `Command.Type` may be good.

madjam:
@kuujo This a very good idea. I too like `Command.Type` and feel it should help the developer express what the command does from the state machine perspective and leave it up to Copycat to interpret what implication it will have on compaction. Also doesn't it give Copycat sufficient info the take care of commit cleaning and not expect state machine to do this part?

kuujo:
I've been trying to think of a way to get rid of `clean()` for all commits. Copycat can handle cleaning commits that are written to a snapshot or are applied and then no longer needed. But the problem that `clean()` solves is determining when a commit that results in long term state is no longer needed. The canonical example of this is a map key. A commit that sets the key `foo` cannot be cleaned until another commit overrides or removes it. So, the map state machine holds the `Commit` object for each key, and once the key is overridden the previous `Commit` value is `clean()`ed.

There's another way to go about handling that relationship, and that is to perform some sort of mapping between commands based on a key defined by the command. This is effectively what Kafka does. Each Kafka message has a key, and during log cleaning it creates a map of keys and retains only the lat instance of each key. The reason I've avoided that approach is because state machines have all the information necessary to do the same thing, and by duplicating that logic internally for log cleaning we'd simply be wasting resources. But I do think that's potentially an abstraction that can be built *on top* of the current `clean`ing API. What is do in that case is provide an extended `Command` interface called `KeyedCommand` or something like that. `KeyedCommand` would add a method to return an object key, and that key could be used by the state machine to handle cleaning of all commits internally. While that would help simplify log cleaning even more, there is some added complexity in the fact that the matching of keys has to reflect contributing to the state machine's state. For a map state machine this is simple to get right, but for other types of state machines it may not be as easy as grouping a set of commands together.

So, that would be a really easy and obvious abstraction to add as well. I like that. But I still wonder if there's an even more elegant way to express when a command that is stored in the state machine for a long period of time is no longer needed. 

Also, there's one more thing I realized here. The `SNAPSHOT` command concept no longer fits with the idea of `Command.Type`. Ultimately, each command is either compacted by snapshots or compacted by log cleaning. So, maybe `Command` needs to have a some way to identify a command as compacted via `SNAPSHOT` or `COMPACT` *and* a way for `COMPACT` commands to specify the command type. There are many potential ways to go about this. Several `Command` interfaces or more enums. Maybe a `SnapshottableCommand` and `CleanableCommand` interface would be nice since they can be used to validate that a state machine supports the command types it expects (i.e. implements `Snapshottable`) at startup.

PR #87, opened 2015-12-15
Simplified LongState + Added unit test for a scenario previously broken

madjam:
Previously the following scenario was broken:

`long.incrementAndGet()
long.set(10)
long.incrementAndGet()`

The above sequence of commands would return `12` instead of the expected value of `11`. This is because the `diff` variable in `LongState` is not reset when the value is set (in `ValueState`).

This commit is an attempt to simplify `LongState` by just keeping track of the current value of the counter.
The new snapshot capability makes this so much simpler.

Also in this commit are changes to make `ValueCommands` more type safe.

### DIFF
--- a/variables/src/main/java/io/atomix/variables/DistributedLong.java
+++ b/variables/src/main/java/io/atomix/variables/DistributedLong.java
@@ -20,6 +20,7 @@ import io.atomix.resource.ResourceType;
 import io.atomix.resource.ResourceTypeInfo;
 import io.atomix.variables.state.LongCommands;
 import io.atomix.variables.state.LongState;
+import io.atomix.variables.state.ValueCommands;
 
 import java.util.concurrent.CompletableFuture;
 
@@ -43,7 +44,7 @@ public class DistributedLong extends AbstractDistributedValue<DistributedLong, L
 
   @Override
   public CompletableFuture<Long> get() {
-    return super.get().thenApply(l -> l != null ? l : 0);
+    return submit(new ValueCommands.Get<>());
   }
 
   /**
--- a/variables/src/main/java/io/atomix/variables/state/LongState.java
+++ b/variables/src/main/java/io/atomix/variables/state/LongState.java
@@ -46,7 +46,7 @@ public class LongState extends ValueState<Long> implements Snapshottable {
   @Override
   public void set(Commit<ValueCommands.Set<Long>> commit) {
     try {
-      value.set((Long) commit.operation().value());
+      value.set(commit.operation().value());
     } finally {
       commit.close();
     }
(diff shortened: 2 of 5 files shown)


jhalterman:
This is very nice :+1:

kuujo:
Fine with me :-)

I think some serious consideration has to be given to how to implement Atomix state machines now. It definitely seems like there are good use cases for both. Ultimately, snapshots do add additional overhead since entries are still compacted in the same way. But for tiny amounts of state that's irrelevant unless someone is creating a million `DistributedValue`s. But there's also overhead to having to hold `Commit` objects in memory e.g. to clean a large map and to having to retain tombstones in a log. Maybe this should be configurable in the future. Data structures that service a lot of deletes can potentially save disk space from snapshotting more than log cleaning since *delete* entries can be removed during minor compaction.

I think `DistributedLong` and `DistributedValue` can and should probably both just be snapshotted now since they result in tiny snapshots anyways.

Should this PR be refactored to just implement snapshots in `ValueState`?

kuujo:
On a related note, we need to get some good testing of large snapshots to try to figure out what the implications are. The problem in Atomix is that the size of snapshots grows with the number of resources that are created as well, and snapshots are currently blocking in Copycat. They're essentially taken at some specific index in the log, and if a lot of resources have been created that could mean a lot of writing to disk by each resource state machine. But we should definitely explore ways to allow snapshots to be taken asynchronously. Some experimentation will have to inform to what extent Atomix resources should use snapshotting.

madjam:
Regarding introducing snapshots in `ValueState`, my impression was it would come in the way of expiring values set with `ttl > 0`

Regarding the high snapshot overhead for large number of resources, does it make sense to divide the snapshotting activity into smaller blocks. In each block we snapshot a subset of resources. Is this what you mean making the activity asynchronous?

kuujo:
So, from what I understand there are two ways you can go about snapshotting. First is to copy the memory before taking the snapshot. That's what LogCabin effectively does by forking the process. The other option is to do as you said - take snapshots in chunks. I think this could actually be feasible, but it's a little complicated as we have to provide some sort of streaming API and state machines have to understand that streaming API. A snapshot relates to a specific position in the log. If a snapshot is taken after index `100` was applied to the state machine, we know it needs to be installed at index `100` after a crash as well. But if snapshots are allowed to be taken while still applying commands to the state machine, we effectively have to associate each chunk of the snapshot with a different index. That can probably be handled internally but it will require some thought.

Anyways, I don't really want to do anything pre-emptive. We should hold off on extending the snapshot interface until the problem being solved becomes more clear. It might be simple enough to just allow state machines to implement another interface to copy their state before taking a snapshot for now.

Good point about the TTL stuff. TTLs do pose a problem and there's probably not much benefit to snapshotting them. We'd still have to hold the TTL `Commit`s in memory and `close` them once the TTL has expired. The alternative is writing the TTL to the snapshot. The `StateMachine` has a `Clock` you can use to get the current logical time. We would write `ttlTime - currentTime` to the snapshot or something, but that can be thought about later.

kuujo:
:+1:

madjam:
I'll create a separate issue for speeding up snapshot creation and installation and continue this conversation there.

Also, good point about snapshotting TTLs. I haven't considered that. I am going to revisit that later to keep the current implementation *simple*. Also `DistributedValue` serves as the simplest example for log cleaning and `DistributedLog` serves that role for snaphotting. So there is some educational value in leaving them as they are now :)

kuujo:
Ahh good point!

I'm actually super proud of myself for finally thinking of a good use for the state machine `Clock` just now :-)

PR #94, opened 2015-12-30
Update CopycatClient usage for client rewrite

kuujo:
The `CopycatClient` was rewritten to handle a variety of issues. This PR is a first step at taking advantage of the new client features. A not-insignificant additional amount of work will have to go into Atomix to handle the changes in logical session IDs across Copycat client session IDs and ensure operations are resubmitted properly across sessions.


kuujo:
The addition of the `SUSPENDED` state to `CopycatClient` should allow some much more elegant handling of inconsistencies inside resource clients. For instance, a leader election should rescind its leadership when the client goes into the `SUSPENDED` state, indicating it can't communicate with the cluster and so another instance may have been elected.

PR #98, opened 2016-01-04
Use atomix-all in the examples

electrical:
In case people decide to extend an example, lets make sure they have a
good first time experience


jhalterman:
Good stiff :)

PR #104, opened 2016-01-07
Add resource options to DistributedMessageBus

kuujo:
This PR modifies the `DistributedMessageBus` resource to use an `Options` class to specify the message bus server `Address`, presenting a more consistent API.


kuujo:
Resource options have been added to `DistributedMessageBus` separately

PR #105, opened 2016-01-07
Push DistributedTaskQueue tasks to consumers in batches

kuujo:
This PR modifies the `DistributedTaskQueue` to pipeline tasks to consumers to improve performance. Currently, consumers acknowledge and retrieve a task in a single commit. This allows multiple tasks to be pushed to each consumer.


electrical:
Performance wise this didn't make much difference but makes sense to have this anyway.

kuujo:
Irrelevant now

PR #106, opened 2016-01-12
Dynamic replica clusters

kuujo:
This PR adds support for dynamically scaling replicas. Replicas use Copycat's new `Cluster` API to elect a leader and dynamically add/remove active, passive, and reserve members from the cluster. This allows users to construct a cluster of any number of `AtomixReplica`s, and Atomix will coordinate the active Raft members, stateful passive members, and stateless reserve members of the cluster. When a member's availability changes, the leader will promote or demote members to rebalance the cluster. The users defines a `quorumHint` and `backupCount` when constructing an `AtomixReplica`, and those numbers are used to balance the number of Raft and non-Raft nodes in the cluster.

The use case for this is simply being able to construct a cluster of `AtomixReplica`s without having to worry about which nodes are servers and which are clients. Previously, if a user were to embed Atomix in a cluster, they would need to permanently assign the role of replica to 3 or 5 or 7 nodes and client to the rest. Now, all nodes can be replicas.


kuujo:
The [Copycat PR](https://github.com/atomix/copycat/pull/112) needs to be merged before this can be built :-)

PR #109, opened 2016-02-05
Replace leaving members to preserve the quorum size

kuujo:
This PR improves the process of balancing the cluster by replacing active servers prior to them leaving the cluster. Previously, active servers would leave the cluster before it was rebalanced to promote a new server to active. However, that means some fault tolerance is lost during that period. Promoting a replacement prior to leaving the cluster ensures that fault tolerance is never lost during membership changes.


kuujo:
This PR exposed a bug in client communication for replicas. Because replicas' clients communicate directly with the local server and don't connect to any other member in the cluster, and because some replicas can be stateless (`RESERVE`), clients for stateless replicas behave differently than those of stateful replicas. In particular, reserve clients cannot receive `SEQUENTIAL` session events since the events are never received and published by the local server. Thus, `RESERVE` replica clients must connect to a remote server to receive published events. 84f8737 resolves that issue by prioritizing connecting the local server, but allowing connections to remote servers in the event that the local server rejects a connection. https://github.com/atomix/copycat/pull/135 altered `RESERVE` servers in Copycat to reject connection attempts from clients and force them to connect to another server.

PR #110, opened 2016-02-05
DistributedMembershipGroup -> DistributedGroup

jhalterman:


kuujo:
@jhalterman gotta update https://github.com/atomix/atomix/blob/dgroup/coordination/src/main/resources/META-INF/services/io.atomix.resource.Resource

jhalterman:
There we go

PR #130, opened 2016-02-20
Do not close resource creation commit when session is closed

### LINKED ISSUE #129: snapshotting logic in ResourceStateMachine
I noticed the following exception after restarting a node.

```
2016-02-19 19:40:46,656 | WARN  | artition-0-state | ServerStateMachineExecutor      | Operation InstanceCommand[resource=474, command=ResourceCommand[command=io.atomix.variables.state.LongCommands$IncrementAndGet@10a8c8e4, consistency=LINEARIZABLE], consistency=LINEARIZABLE] failed
io.atomix.manager.state.ResourceManagerException: unknown resource: 474
	at io.atomix.manager.state.ResourceManagerState.operateResource(ResourceManagerState.java:99)[68:org.onosproject.onlab-thirdparty:1.5.0.SNAPSHOT]
	at io.atomix.manager.state.ResourceManagerState$$Lambda$52/853973916.apply(Unknown Source)[68:org.onosproject.onlab-thirdparty:1.5.0.SNAPSHOT]
	at io.atomix.copycat.server.state.ServerStateMachineExecutor.executeOperation(ServerStateMachineExecutor.java:143)[68:org.onosproject.onlab-thirdparty:1.5.0.SNAPSHOT]
	at io.atomix.copycat.server.state.ServerStateMachine.executeCommand(ServerStateMachine.java:925)[68:org.onosproject.onlab-thirdparty:1.5.0.SNAPSHOT]
	at io.atomix.copycat.server.state.ServerStateMachine.lambda$apply$29(ServerStateMachine.java:839)[68:org.onosproject.onlab-thirdparty:1.5.0.SNAPSHOT]
```
the key thing is this exception only occurs after a snapshot is taken. If no snapshots were taken you don't see this error.

In `ResourceManagerState`, the snapshot logic is to merely take snapshots of the resource state machines. But `ResourceManagerState` itself has some state that probably needs to be included in the snaphot.

kuujo:
This PR attempts to resolve #129. The problem is, when the first client opens a resource, the client's `GetResource` commit becomes the commit that created that resource. However, `GetResource` is also used as the commit that created a logical session/connection from the client to the replicated resource. When a session is closed or the client closes the resource, the commit that created the session is closed:
https://github.com/atomix/atomix/blob/master/manager/src/main/java/io/atomix/manager/state/ResourceManagerState.java#L322

But the commit that created the *first* session for a resource was also the commit that created the resource. Thus, we resolve this by preventing the `ResourceHolder`'s `commit` from being `close`d by a session being closed.

@madjam can you give this a test and see if it resolves your issue?

### DIFF
--- a/examples/distributed-value/src/main/java/io/atomix/examples/variables/DistributedValueExample.java
+++ b/examples/distributed-value/src/main/java/io/atomix/examples/variables/DistributedValueExample.java
@@ -63,7 +63,7 @@ public class DistributedValueExample {
         .withMinorCompactionInterval(Duration.ofSeconds(30))
         .withMajorCompactionInterval(Duration.ofMinutes(1))
         .withMaxSegmentSize(1024 * 1024 * 8)
-        .withMaxEntriesPerSegment(1024 * 1024)
+        .withMaxEntriesPerSegment(1024 * 8)
         .build())
       .build();
 
--- a/examples/distributed-value/src/main/resources/logback.xml
+++ b/examples/distributed-value/src/main/resources/logback.xml
@@ -23,7 +23,7 @@
 
   <logger name="io.atomix.catalyst" level="WARN" />
 
-  <root level="DEBUG">
+  <root level="INFO">
     <appender-ref ref="STDOUT" />
   </root>
 </configuration>
--- a/manager/src/main/java/io/atomix/manager/state/ResourceManagerState.java
+++ b/manager/src/main/java/io/atomix/manager/state/ResourceManagerState.java
@@ -230,7 +230,10 @@ public class ResourceManagerState extends StateMachine implements SessionListene
         resourceHolder.stateMachine.unregister(sessionHolder.session);
         resourceHolder.stateMachine.close(sessionHolder.session);
       } finally {
-        sessionHolder.commit.close();
+        // Ensure that the commit that created the resource is not closed even of the client closed its reference to it.
+        if (sessionHolder.commit != resourceHolder.commit) {
+          sessionHolder.commit.close();
+        }
       }
     }
   }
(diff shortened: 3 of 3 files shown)


kuujo:
Ugh... sloppy me I got some of the changes I made when I was trying to reproduce this in here :-( haha

kuujo:
I suppose the more elegant way to do this would be to separate commands for creating a resource and opening a session to a resource, but combining them in a single command prevents race conditions. We could do some optimistic locking to open a session to a resource, though. Resource creation on a client and deletion of resources should be generally infrequent operations, so that may be fine. I may put something like that together since it should make for cleaner interaction with the state machine.

This smells funny.

madjam:
:+1: 
This takes care of it.

I agree separate APIs for creation and access makes it cleaner. Also enables an administrator to implement access control easily. Having said that for some users (like me right now) the current behavior simplifies their code :)

kuujo:
The way I'll implement this is just as some cleanup for the code smell in the resource manager state machine. From the user's perspective, nothing will change. I'm really just separating a single state machine method into two. A call to `getValue` will just submit a `CreateResource` and then `OpenResource` command. `CreateResource` will ensure the resource has been created and return the resource ID which can be used by the `OpenResource` command to open a new logical session.

I'm going to close this PR and resubmit a new one with the changes.

kuujo:
Actually, I decided to keep this open and get it merged. I think the addition of separate commands for creating and opening a resource introduces too many race conditions for my liking at this point. In the interest of stability, having a single atomic operation to open a resource is safest. If there are separate *create* and *open* commands, we'll have to account for cases where a resource is created and then deleted before it's opened by the client. That may be rare, but it requires a lot more client-side logic and exceptions to handle gracefully, and the single atomic operation doesn't have those issues.

jhalterman:
:+1:  to tackling separate commands later.

PR #135, opened 2016-02-28
Group consistent hashing/partitions

kuujo:
This PR adds helper methods to the `DistributedGroup` resource to aid in consistent hashing and partitioning of resources. Groups can now be configured when created, and the group will manage partitions within the state machine using a consistent hashing algorithm:

```java
DistributedGroup.Config config = DistributedGroup.config()
  .withPartitions(32)
  .withVirtualNodes(100)
  .withReplicationFactor(3);

DistributedGroup.Options options = DistributedGroup.options()
  .withPartitioner(HashCodePartitioner.class);

DistributedGroup group = atomix.getGroup("foo", config, options);
```

Each partition is hashed to a point on the ring, and the `n` replicas following that point on the ring are listed in the `Partition.members` list, as is described in the Dynamo paper. On the client side, a `Partitioner` maps arbitrary `Object`s to partitions.

```java
Partition partition = group.partition("foo");
partition.members().forEach(member -> {
  member.send("foo");
});
```

Partitions can also be accessed on a per-member basis:

```java
group.join(member -> {
  member.partitions().forEach(partition -> {
    System.out.println(partition.id());
  });
});
```

This is ultimately the beginning of introducing different replication algorithms into Atomix. Eventually, the `ResourceStateMachine` will be abstracted into a more agnostic framework, and state machines/services that *can* be partitioned will be manageable *either* through consensus *or* consistent hashing.


kuujo:
I'm going to close this pull request. I feel the feature is useful, and much of the code can be ported over to a new pull request. But some work needs to be done on configuration management to make this practical. Namely, configurations need to be static once a resource is created, and configurations need to be pushed to clients to allow clients to do consistent hashing rather than doing it in the state machine.

PR #140, opened 2016-03-02
Refactor Resource into an interface and add factories

kuujo:
This PR refactors the `Resource` class into an interface and adds a `ResourceFactory` for creating resource instances. Additionally, resources are now configured in the `ResourceStateMachine` constructor to ensure configurations remain static.


kuujo:
Documentation will be improved in a separate pull request.

PR #147, opened 2016-03-05
Create standalone resource server.

jhalterman:
Create standalone resource server.
Move properties from atomix module to atomix-resource-manager.
Remove standalone server example.
properties -> options class naming. This reflects the idea that config can potentially (in the future) be populated by various sources, including properties files.
Use Address host string constructor
Move options to .options packages


kuujo:
Some changes just got merged into Copycat to rename client/server start methods and change how servers join the cluster. I think that impacts this PR a bit.

PR #176, opened 2016-04-22
enable submitting of distributed group metadata

davidmoravek:
Hi, first of all, thanks for an amazing job you're doing with atomix!

We're currently trying to adopt it in a new project and we run into usability issues, when it comes to service discovery. I think it would be really helpful if we could submit metadata, when joining a distributed group. Metadata can contain additional info about node, e.g. host:port of some RPC and so on...

I made a POC, can you please go through it and let me know if this is a good approach?

Thanks,
D.

### DIFF
--- a/groups/src/main/java/io/atomix/group/DistributedGroup.java
+++ b/groups/src/main/java/io/atomix/group/DistributedGroup.java
@@ -481,6 +481,8 @@ public interface DistributedGroup extends Resource<DistributedGroup> {
    */
   CompletableFuture<LocalMember> join(String memberId);
 
+  CompletableFuture<LocalMember> join(String memberId, Object metadata);
+
   /**
    * Adds a listener for members joining the group.
    * <p>
--- a/groups/src/main/java/io/atomix/group/GroupMember.java
+++ b/groups/src/main/java/io/atomix/group/GroupMember.java
@@ -17,6 +17,8 @@ package io.atomix.group;
 
 import io.atomix.group.messaging.MessageClient;
 
+import java.util.Optional;
+
 /**
  * A {@link DistributedGroup} member representing a member of the group controlled by a local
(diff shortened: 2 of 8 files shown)


kuujo:
Totally agree. If there are usability issues we can fix them quickly :-)

At a quick glance, I think this approach looks fine. Let me mull it over a bit and I'll take a look at it more closely this evening.

Thanks!

kuujo:
So, for a bit of history on this as well, the original implementation of ‘DistributedGroup` actually had a properties-like object associated with each group and member. It was removed since technically it didn't add anything of value. You can create e.g. a map and store that type of information.

I think the only way this makes real sense for service discovery is allowing metadata to be passed when joining the group as this PR does so joining the group with metadata is an atomic operation.

I think this is good. The only part I'm unsure about is adding a generic type to `DistributedGroup` since the metadata is really optional and based on how the group is used. I might rather just make the `metadata()` method return type generic.

kuujo:
BTW I like the use of `Optional` but it's inconsistent with the current API. That's not an issue with this PR but rather an issue with the existing APIs. I'm thinking maybe I should submit a separate PR adding `Optional` to all the existing interfaces where it makes sense and we can merge both at the same time.

jhalterman:
I would probably agree with @kuujo's comment regarding the type param. I don't think the metadata's type adds enough value to all that a `DistributedGroup` or `GroupMember` does to warrant being a class-level type param.

davidmoravek:
I reckon, leaving out the class level generic type, would simplify thinks a lot.  And since the metadata are just optional, it doesn't really add much of a value. I'll try to rework it.

davidmoravek:
Simplified, can you please review again? Thanks

D.

kuujo:
👍

PR #180, opened 2016-05-05
Allow specifying ConnectionStrategy for AtomixClient

mnicky:
I've run into the case when I want to specify my own ConnectionStrategy for AtomixClient, but it wasn't exposed in the API. So I added required methods to the respective builders.


kuujo:
Good with me!
That failure is from a missing snapshot. I bumped the dependency version separately.

PR #189, opened 2016-06-16
Use expire() method to more accurately test resource recovery

kuujo:
This PR fixes a bug in the resource recovery test. Because of how the test was written, clients were not transitioning states in the same way they would actually transition in practice, forcing the `RecoveryStrategy` to be called. We use the new `ClientSession.expire()` method to expire the session and ensure client states are consistent with practice during testing.


kuujo:
Tests will fail here until atomix/copycat#234 is merged

madjam:
👍

kuujo:
Tests are good.

PR #190, opened 2016-06-17
WIP: Distributed group recovery

dmvk:
Lets address some of the atomix client consistency guarantees. Let's take DistributedGroup for example. I create a new group and join it. Now the client's session expires for some reason. Default recoveryStrategy for atomix client is set to RECOVERY (cannot be overridden right now).

Now when client successfully recovers, it gets a brand new session id. The problem is, it is not a member of distributed group anymore.

Another problem is, that members of distributed group are cached. It does sync() when resource opens, but then updates its state based on onJoin and onLeave events. So with new session ID it doesn't receive those on reconnect and doesn't do any "resync" either.

I think that atomix client should be able to handle this. Unfortunately I'm not sure how to properly address this, so at least I'm adding a test case to replicate the issue.

What you guys think?

Thanks,
D.

### DIFF
--- a/groups/src/main/java/io/atomix/group/internal/MembershipGroup.java
+++ b/groups/src/main/java/io/atomix/group/internal/MembershipGroup.java
@@ -60,6 +60,7 @@ public class MembershipGroup extends AbstractResource<DistributedGroup> implemen
   private final Map<String, AbstractGroupMember> members = new ConcurrentHashMap<>();
   private final MessageProducerService producerService;
   private final MessageConsumerService consumerService;
+  private final Map<String, GroupCommands.Join> joinCommands = new ConcurrentHashMap<>();
 
   public MembershipGroup(CopycatClient client, Properties options) {
     super(client, new ResourceType(DistributedGroup.class), options);
@@ -129,10 +130,12 @@ public class MembershipGroup extends AbstractResource<DistributedGroup> implemen
   private CompletableFuture<LocalMember> join(String memberId, boolean persistent, Object metadata) {
     // When joining a group, the join request is guaranteed to complete prior to the join
     // event being received.
-    return client.submit(new GroupCommands.Join(memberId, persistent, metadata)).thenApply(info -> {
+    final GroupCommands.Join cmd = new GroupCommands.Join(memberId, persistent, metadata);
+    return client.submit(cmd).thenApply(info -> {
       AbstractGroupMember member = members.get(info.memberId());
       if (member == null || !(member instanceof LocalGroupMember)) {
         member = new LocalGroupMember(info, this, producerService, consumerService);
+        joinCommands.put(memberId, cmd);
         members.put(info.memberId(), member);
       }
       return (LocalGroupMember) member;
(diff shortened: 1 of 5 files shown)


JozoVilcek:
I am affected by this issue as well. Looking forward for the solution

davidmoravek:
Can you guys please provide any feedback? I'd be really helpful ;) Thanks

D.

kuujo:
Thanks for the reminder! Totally got buried in my email. I shouldn't procrastinate...

kuujo:
Okay, so you're totally right. A resource that stores state on the client needs to recover it when a session is recovered.

The way to do this would be by listening for state changes in the resource's `CopycatClient`. In the `DistributedGroup.open` method we should register a client state change listener. When the client transitions to `SUSPENDED` state and back to `CONNECTED` the resource can ensure any local members are registered in the group with an idempotent join. A client can transition to `SUSPENDED` and not lose its session, so the resource probably needs to have a UUID or something like that to ensure a resource doesn't join the group multiple times. Alternatively, you can listen to the `client.session()` state to determine when the session is *definitely* expired by the `Session.State.EXPIRED`, in which case any new join does not need to be idempotent.

The problem is, messages sent to the member can still be lost unless it uses a persistent member ID. If an explicit member ID is provided at join, messages and other state associated with the member is preserved I the state machine until an explicit `leave`. Whether a member should always have an explicit ID can be debated. The problem with generating one would be deleting state of the node actually crashes and that member ID is gone.

There are definitely limitations to what we can do here. It's impossible to get around session expirations without losing some consistency by their nature. But for persistent members, we can certainly prevent them from being removed from the group and allow recovered sessions to recover the same group member. This is already built in to the state machine, so it should be as simple as rejoining the group with the same member ID. But we have to ensure that state is removed from the state machine when necessary. This may mean that only members with explicit member IDs can be automatically recovered and the loss of a session for a member without an explicit ID results in a loss of the resource instance - whatever that means - forcing the user to rejoin the group explicitly.

dmvk:
Hi, sorry for late reply, I was on a vacation. I agree that possibilities are very limited when it comes to session expiration. Still I think, it should be handled transparently by client. So I propose adding `onRecovery` method to `Resource` interface, which lets you know, that you're starting from "fresh" state and you're application should handle this if necessary. I've tried to create initial implementation for `AbstractResource` and `MembershipGroup`.

Also, I needed to fix State transitions in Copycat client in order to implement this.

https://github.com/atomix/copycat/pull/240

Do you think this is a good approach?

Thanks,
D.

dmvk:
ping, thanks ;)

D.

kuujo:
I think that makes sense

kuujo:
This looks fine but I need to find a computer to look at the tests :-)

kuujo:
This is the failure that's occurring in Travis:
```
testRecovery(io.atomix.group.DistributedGroupTest)  Time elapsed: 9.225 sec  <<< FAILURE!
java.lang.AssertionError: expected:<1> but was:<2>
	at io.atomix.group.DistributedGroupTest.testRecovery(DistributedGroupTest.java:652)
```

I'll look in to it if I get a chance tonight

JozoVilcek:
Hi guys, any progress and good news on this?

kuujo:
This looks good to me. I'm going to fix the tests separately.

PR #198, opened 2016-07-21
Make Map, Set, MultiMap and Value commands with time-to-live to have EXPIRING compaction mode

leventov:
… instead of SEQUENTIAL, as suggested in the Copycat documentation


kuujo:
Thanks. This is correct. Restarting tests, but the failures are unrelated.

PR #204, opened 2016-08-30
Add ability to get resource server

mnicky:
I have a use case where I need an ability to get the `ResourceServer` used by `AtomixReplica` to be able to register `onStateChange` listener to the underlying `CopycatServer` instance.


dmvk:
Hi guys! I think, this would be really useful. Is there any reason it hasn't been merged? (Travis fails on a flipping test, that has nothing to do with this PR)

Thanks,
D.

PR #208, opened 2016-12-26
Group test fixes

kuujo:
This PR fixes some issues in group tests wherein the tests don't properly await events received on the clients. There seem to be some other issues in group tests that only show up in Travis, so I'll address those here in the test runs.


kuujo:
This PR also resolves a bug in the `DistributedGroup` recovery method to ensure that local group members are consistent with the cluster prior to new join events being handled.

PR #213, opened 2017-01-11
Ensure producer message ID is incremented when asynchronous message is sent

kuujo:
This PR fixes a bug in the base group message producer. `messageId` was not being incremented when `ASYNC` messages were sent.


jhalterman:
👍

PR #228, opened 2017-02-01
Clean up resource manager state machine

kuujo:
This PR is some minor cleanup of the `ResourceManagerStateMachine`. It removes an unused internal class and ensures that `GetResource` commits are properly compacted from the Raft log.


jhalterman:
👍

Issue #243, opened 2017-03-21
ConcurrentModificationException in ONOS shutdown


kalagesa:
Hi,

We observerd ConcurrentModificationException  in ONOS shutdown. We have also raised a JIRA for the same in ONOS [ONOS-6148]. 
We analysed the stack trace and identified that _instances_ map which is a hashMap in ResourceClient.java is causing the problem. In ResourceClient.close() method we have a map iterator.  While iterating over the map, resources are removed from _instances_ map in another method **private synchronized void close(ResourceInstance instance)**. This is causing ConcurrentModificationException.
We can avoid this exception by changing _instances_ to ConcurrentHashMap.

Thanks,
Kavitha

kuujo:
I'm a little confused by why there's a `ConcurrentModificationException` considering all the access to the map is synchronized inside `ResourceClient`, but I don't see any reason we can't just change it and forget about it.

PR #249, opened 2017-04-13
Fix for ConcurrentModification Exception.Issue #243

deepavad:


kuujo:
The problem here is going to be if an individual instance is closed then it's not removed from the `instances` map any more. That means a call to `getResource` can then return a closed instance.

Last night I just changed the `instances` and `futures` map to a `ConcurrentHashMap`: #248
That should resolve the issue

PR #253, opened 2017-05-02
Bump Copycat dependency to 1.2.6

kuujo:


kuujo:
I'm not sure tests will pass. Maven central sync seems to be faster these days, but probably not _that_ fast.

kuujo:
Well that was fast...

PR #254, opened 2017-05-08
Bump Copycat dependency version to 1.2.7

kuujo:
This isn't actually in Central yet, but should be in a few. I'll restart the tests when it gets there.


kuujo:
Restarting tests

PR #259, opened 2017-07-27
Add leadership transfer functionality

kuujo:
This PR adds support for transferring leadership to a specific node. This is done by simply calling `promote()` on the `RaftServer` to which to transfer leadership. Leadership transfer is implemented by sending a `TransferRequest` to the leader which will then block writes and replicate the rest of its log to the follower before stepping down and responding to the request with a `TransferResponse`. Once the follower receives the response, it immediately transitions to the `CANDIDATE` role to get elected. If for some reason the follower is not the next node elected, the transfer will fail.

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftContext.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftContext.java
@@ -15,6 +15,7 @@
  */
 package io.atomix.protocols.raft.impl;
 
+import io.atomix.protocols.raft.RaftException;
 import io.atomix.protocols.raft.RaftServer;
 import io.atomix.protocols.raft.cluster.MemberId;
 import io.atomix.protocols.raft.cluster.RaftMember;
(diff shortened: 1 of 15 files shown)


protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftContext.java
```diff
@@ -643,6 +647,54 @@ private void unregisterHandlers(RaftServerProtocol protocol) {
...
+      // state and attempt to get elected.
+      RaftMember member = getCluster().getMember();
+      RaftMember leader = getLeader();
+      if (leader != null) {
+        protocol.transfer(leader.memberId(), TransferRequest.newBuilder()
+            .withMember(member.memberId())
+            .build()).whenComplete((response, error) -> {
+          if (error != null) {
+            future.completeExceptionally(error);
+          } else if (response.status() != RaftResponse.Status.ERROR) {
```
jhall11: should this be: if status() == ERROR?

kuujo: Oops it seems my update to the commit didn't make it to GitHub one sec

Issue #265, opened 2017-08-29
Find a log compaction bug in RaftServiceManager with 2.0 final edition.


StackWu:
Hi, my friends ! I'm very interested in your project which is the best implementation of the raft consensus algorithm in JAVA. And the project give me a lot of help to my work.

Nevertheless, I find a bug about the log compaction. The compactLog method will run only one time after I start the server. Then it won't run anymore.

I guess the reason is that you make a mistake in the schedule task(The code is in the constructor method of the  RaftServiceManager ). Here is your original code 

`threadContext.schedule(Duration.ofMillis(COMPACT_INTERVAL_MILLIS), this::compactLog);`

And I try to rewrite it 

 `threadContext.schedule(Duration.ZERO, Duration.ofMillis(COMPACT_INTERVAL_MILLIS), this::compactLog);`

I don't know whether I misunderstood what you meant about the code. But it's so important that I can't ignore it. The rapidly growing logs make me uncomfortable. Please help me to solve the problem. Thanks very much !

By the way, I miss the ability to control the compact interval because I could readjust it to adapt environment. Now the compact interval is hardcoded. Maybe you have a better reason. Could you share with me something about that? Thanks Again.

kuujo:
It appears you're right about the compaction scheduler. It will be fixed and released today.

As for the frequency of compaction, it's hard coded at 10 seconds because compaction only occurs after the log rolls over to a new segment. Unless you need the log to be compacted more frequently than every ten seconds, there's not really good reason to make the time configurable. The majority of checks will be no-ops. When the timer expires, we check to see if the log is "compactable" meaning it has segments that can be removed from disk. Only if it's compactable will snapshots be taken. We do this since deleting a segment from disk is much more efficient than rewriting the log with the most recent entries. So, if you want to control the frequency of log compaction you should change the size of the log segments.

kuujo:
Fixed

PR #266, opened 2017-08-29
Ensure log compaction is rescheduled after completion

kuujo:
#265


protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
```diff
@@ -494,6 +501,7 @@ private void compactLog() {
           .whenCompleteAsync((result, error) -> {
             logger.info("Compacting logs up to index {}", lastApplied);
             log.compact(lastApplied);
+            scheduleCompaction();
```
jhalterman: Can log.compact ever throw? Want to make sure we don't miss the next `scheduleCompaction`.

kuujo: I suppose it could, due to an `IOException` in the log. Should probably log that as an `ERROR`

PR #268, opened 2017-08-29
Disable leadership transfer test due to non-determinism

kuujo:


kuujo:
Let's see how this test does

kuujo:
🤔

PR #273, opened 2017-09-05
Snapshot replication fixes

kuujo:
This PR fixes bugs in the snapshot replication protocol that were exposed after fixing #265. It makes minor changes to the way snapshots are replicated and installed to avoid certain race conditions, and it improves compaction scheduling to reduce the chance snapshots need to be replicated to followers.


protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
```diff
@@ -457,21 +530,40 @@ else if (nextIndex == index) {
...
+   * Schedules a log compaction.
+   */
+  private void scheduleCompaction(long lastApplied) {
+    // Wait half the time since the last compaction before compacting the logs to ensure followers can keep
+    // up with the leader without replicating snapshots.
+    long compactTime = System.currentTimeMillis();
+    long lastCompactTime = this.compactTime;
+    this.compactTime = compactTime;
+    if (lastCompactTime > 0) {
+      Duration duration = Duration.ofMillis((compactTime - lastCompactTime) / 2);
```
jhall11: Do we want to set a minimum and/or maximum for this duration? Could we get into a state where compaction was held up for a long time? It could take several rounds to get back to the normal interval

kuujo: Hmm... this can probably just be set to a static time for now. We can reasonably expect followers to stay within x seconds of the leader.

kuujo: Fixed

kuujo:
I think this is just about g2g

PR #276, opened 2017-09-12
Clean up log compaction/snapshot replication protocol

kuujo:
This PR is a major cleanup of the log compaction/snapshot replication protocol and various related protocols. The changes in this PR are aimed at reducing the frequency of timeouts in large clusters. I've consolidated all of the commits into a single PR for expediency, but each individual commit could legitimately be its own PR. I'd recommend reviewing it one commit at a time. The commit messages should describe the changes made within the commit.

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/roles/LeaderAppender.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/roles/LeaderAppender.java
@@ -255,7 +255,7 @@ final class LeaderAppender extends AbstractAppender {
       TimestampedFuture<Long> future = iterator.next();
 
       // If the future is timestamped prior to the last heartbeat to a majority of the cluster, complete the future.
-      if (future.timestamp <= heartbeatTime) {
+      if (future.timestamp < heartbeatTime) {
         future.complete(null);
         iterator.remove();
       }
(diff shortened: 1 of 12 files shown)


protocols/raft/src/main/java/io/atomix/protocols/raft/roles/LeaderAppender.java
```diff
@@ -450,7 +450,7 @@ protected void failAttempt(RaftMemberContext member, RaftRequest request, Throwa
     // Verify that the leader has contacted a majority of the cluster within the last two election timeouts.
     // If the leader is not able to contact a majority of the cluster within two election timeouts, assume
     // that a partition occurred and transition back to the FOLLOWER state.
-    if (System.currentTimeMillis() - Math.max(getHeartbeatTime(), leaderTime) > raft.getElectionTimeout().toMillis() * 2) {
+    if (member.getFailureCount() >= 3 && System.currentTimeMillis() - Math.max(getHeartbeatTime(), leaderTime) > raft.getElectionTimeout().toMillis() * 2) {
```
jhalterman: Should move this magic number to a constant.

kuujo: hmm actually `raft.getElectionTimeout().toMillis()` should also be using the existing field that already has the milliseconds computed

PR #278, opened 2017-09-12
Add session manager tests

kuujo:
Add simple tests for the `RaftSessionManager`


jhall11:
Why does coverage decrease when adding a test?

kuujo:
Hahaha that's odd

Issue #233, opened 2017-03-12
Fix BalancingClusterManager


kuujo:
`BalancingClusterManager` has been unmaintained and currently does not work as advertised. Ensure the cluster manager works as expected and add tests.

Issue #210, opened 2017-01-05
AtomixReplica with BalancingClusterManager hangs at startup 


santospm:
I may've found a bug in BalancingClusterManager; using it causes AtomixReplica to hang.  I traced the cause the 'start' method; it returns a "null" CompletableFuture (io.atomix.cluster.BalancingClusterManager:line 75).  My workaround was to create a subclass and override and return a non null CompletableFuture:
AtomixReplica.builder(...)
            ...
                .withClusterManager(new BalancingClusterManager(3, 1) {
                    @Override
                    public CompletableFuture<Void> start(Cluster cluster, AtomixReplica replica) {
                        super.start(cluster, replica);
                        return CompletableFuture.completedFuture(null);
                    }
                })
                .build();

PR #283, opened 2017-10-08
Fix balancing cluster manager and adds test cases

rohangulati:
Fixes https://github.com/atomix/atomix/issues/233

Addresses the following issue
* Fixes null pointer issue mentioned in https://github.com/atomix/atomix/issues/210
* Adds test cases for `BalancingClusterManager`

### DIFF
--- a/core/src/test/java/io/atomix/cluster/BalancingClusterManagerTest.java
+++ b/core/src/test/java/io/atomix/cluster/BalancingClusterManagerTest.java
@@ -1,0 +1,299 @@
+package io.atomix.cluster;
+
+import io.atomix.AtomixReplica;
+import io.atomix.catalyst.transport.Address;
+import io.atomix.catalyst.transport.local.LocalServerRegistry;
+import io.atomix.catalyst.transport.local.LocalTransport;
+import io.atomix.copycat.server.cluster.Cluster;
+import io.atomix.copycat.server.cluster.Member;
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/atomix/cluster/BalancingClusterManagerTest.java
```diff
@@ -0,0 +1,299 @@
...
+        if (type == Member.Type.PASSIVE) {
+          resume();
+        }
+      });
+    });
+
+    // adding a RESERVE member will result in promotion of this member to PASSIVE
+    AtomixReplica joiner = createReplica(nextAddress(), AtomixReplica.Type.RESERVE, clusterManager);
+    joiner.join(this.members).thenRun(this::resume);
+    await(10000, 2);
```
johnou: Hard to see because I'm reviewing by mobile but where is await defined? Can we use a count down latch or join on the cf here instead?

jhalterman: It's inherited from `ConcurrentTestCase`. It blocks the main test thread until some other thread calls `this::resume`.

johnou: Okay fine as is then.

johnou:
Lgtm

johnou:
@kuujo good to merge?

PR #285, opened 2017-10-13
Replicate from leader to passive nodes

kuujo:
This PR changes replication to passive nodes to be done by leaders. Previously, followers replicated to passive nodes. But the removal of node statuses from the Raft protocol also removed the ability to balance passive replication. Leaders need to replicate directly to followers.


jhall11:
I have some reservations about the scalability of this, but I think in larger cluster sizes, this probably isn't the biggest scalability issues

PR #291, opened 2017-10-31
Remove RESERVE role and block joining servers until caught up to the leader

kuujo:
This PR removes the `RESERVE` role from Raft servers which will be replaced by simply shutting down servers.

It also ensures that servers joining the cluster as `ACTIVE` members are blocked from being added to the quorum until caught up to the leader. This ensures that e.g. a node being added to an existing 1-node cluster won't be added with an empty log, thus blocking writes until it's caught up to the leader.


protocols/raft/src/main/java/io/atomix/protocols/raft/RaftServer.java
```diff
@@ -575,17 +571,6 @@ public Builder withName(String name) {
     }
 
     /**
-     * Sets the initial server member type.
-     *
-     * @param type The initial server member type.
-     * @return The server builder.
-     */
-    public Builder withType(RaftMember.Type type) {
```
kuujo: This method should be deprecated

PR #292, opened 2017-10-31
Reset known leader when command fails due to ConnectException

kuujo:
This PR ensures that Raft proxies locate new leaders more quickly when a leader node crashes. Currently, when a leader crashes the proxy will continue attempting to submit commands to that leader until the next keep-alive resets the leader info. This can result in timeouts when queries are blocked waiting for commands to be completed by sequence number. But this PR forgets the leader when a `ConnectException` is returned by a command so that the client will proxy commands through any node until the new leader is discovered.

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/impl/DefaultRaftMetadataClient.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/impl/DefaultRaftMetadataClient.java
@@ -61,8 +61,8 @@ public class DefaultRaftMetadataClient implements RaftMetadataClient {
   }
 
   @Override
-  public Collection<MemberId> getServers() {
-    return selectorManager.servers();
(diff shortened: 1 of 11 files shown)


protocols/raft/src/main/java/io/atomix/protocols/raft/impl/DefaultRaftMetadataClient.java
```diff
@@ -62,7 +62,7 @@ public MemberId getLeader() {
 
   @Override
   public Collection<MemberId> getServers() {
-    return selectorManager.servers();
+    return selectorManager.members();
```
johnou: @kuujo was this intentional or should the getter be renamed too?

kuujo: hmm yeah missed a rename there

PR #294, opened 2017-11-01
Use phi accrual failure detectors for Raft elections and session timeouts

kuujo:
This PR refactors how leadership elections and session expirations are handled in Raft.

It adds a phi accrual failure detector used to determine when to start a new election. In order to avoid multiple servers starting an election at the same time, randomized timers are used to check the current phi value. The Raft election timeout is used as a fallback to ensure the timeout doesn't surpass that point.

Sessions are also expired using phi accrual failure detectors. This is done by the leader sending heartbeats to clients. New sessions are opened with a minimum timeout, and the leader sends heartbeats to the clients at the rate of the minimum session timeout. Sending heartbeats from the leader to clients also ensures clients resolve new leaders as soon as possible. In order to account for the time period during which an old leader crash was being detected and a new leader was being elected, nodes track the last heartbeat time and subtract that time from session timeouts. This means sessions can be expired via the failure detector immediately after a leader change if the client can't be reached by the leader.

### DIFF
--- a/protocols/failure-detection/src/main/java/io/atomix/protocols/phi/PhiAccrualFailureDetector.java
+++ b/protocols/failure-detection/src/main/java/io/atomix/protocols/phi/PhiAccrualFailureDetector.java
@@ -18,7 +18,6 @@ package io.atomix.protocols.phi;
 import org.apache.commons.math3.stat.descriptive.DescriptiveStatistics;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
  * Phi Accrual failure detector.
@@ -27,6 +26,15 @@ import static com.google.common.base.Preconditions.checkNotNull;
  */
 public class PhiAccrualFailureDetector {
 
+  /**
+   * Returns a new failure detector builder.
+   *
+   * @return a new failure detector builder
+   */
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
   // Default value
   private static final int DEFAULT_WINDOW_SIZE = 250;
   private static final int DEFAULT_MIN_SAMPLES = 25;
@@ -34,13 +42,13 @@ public class PhiAccrualFailureDetector {
 
   private final int minSamples;
   private final double phiFactor;
-  private final History history = new History();
+  private final History history;
 
   /**
    * Creates a new failure detector with the default configuration.
    */
   public PhiAccrualFailureDetector() {
-    this(DEFAULT_MIN_SAMPLES, DEFAULT_PHI_FACTOR);
+    this(DEFAULT_MIN_SAMPLES, DEFAULT_PHI_FACTOR, DEFAULT_WINDOW_SIZE);
   }
 
   /**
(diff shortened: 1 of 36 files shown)


protocols/failure-detection/src/main/java/io/atomix/protocols/phi/PhiAccrualFailureDetector.java
```diff
@@ -29,93 +25,69 @@
  * <p>
  * Based on a paper titled: "The φ Accrual Failure Detector" by Hayashibara, et al.
  */
-public class PhiAccrualFailureDetector<T extends Identifier> {
-  private final Map<T, History> states = Maps.newConcurrentMap();
+public class PhiAccrualFailureDetector {
```
kuujo: We should add a builder for this class

protocols/failure-detection/src/main/java/io/atomix/protocols/phi/PhiAccrualFailureDetector.java
```diff
@@ -29,93 +25,69 @@
...
   /**
    * Report a new heart beat for the specified node id.
    *
-   * @param nodeId      node id
    * @param arrivalTime arrival time
    */
-  public void report(T nodeId, long arrivalTime) {
-    checkNotNull(nodeId, "NodeId must not be null");
+  public void report(long arrivalTime) {
+    checkNotNull("NodeId must not be null");
```
kuujo: This assertion is checking a string literal

protocols/raft/src/main/java/io/atomix/protocols/raft/roles/FollowerRole.java
```diff
@@ -59,39 +55,38 @@ public FollowerRole(RaftContext context) {
...
+    return super.open().thenRun(this::startHeartbeatTimer).thenApply(v -> this);
   }
 
   /**
    * Starts the heartbeat timer.
    */
-  private void startHeartbeatTimeout() {
+  private void startHeartbeatTimer() {
     log.trace("Starting heartbeat timer");
+    AtomicLong lastHeartbeat = new AtomicLong();
```
kuujo: This value needs to be updated, otherwise the last heartbeat time is always reported.

protocols/raft/src/main/java/io/atomix/protocols/raft/roles/LeaderRole.java
```diff
@@ -179,6 +186,106 @@ private void appendMembers() {
...
+  private void resetHeartbeatTimer(MemberId member) {
+    // Compute the smallest timeout of all open sessions for the member.
+    OptionalLong minTimeout = raft.getSessions().getSessions().stream()
+        .filter(s -> s.memberId().equals(member))
+        .mapToLong(s -> s.minTimeout())
+        .min();
+
+    Scheduled oldTimer;
+    if (minTimeout.isPresent()) {
+      Scheduled newTimer = raft.getThreadContext().schedule(
```
kuujo: It sure if we should be using a recurring timer or setting a timer recursively when responses are received.

protocols/raft/src/main/java/io/atomix/protocols/raft/roles/LeaderRole.java
```diff
@@ -179,6 +186,106 @@ private void appendMembers() {
...
+   */
+  private void sendHeartbeat(MemberId member, Collection<RaftSessionContext> sessions) {
+    HeartbeatRequest request = HeartbeatRequest.newBuilder()
+        .withLeader(raft.getCluster().getMember().memberId())
+        .withMembers(raft.getCluster().getMembers().stream()
+            .map(RaftMember::memberId)
+            .filter(m -> m != null)
+            .collect(Collectors.toList()))
+        .build();
+    log.trace("Sending {}", request);
```
kuujo: Add the `MemberId` to this log statement

protocols/raft/src/main/java/io/atomix/protocols/raft/session/impl/RaftSessionContext.java
```diff
@@ -52,19 +53,23 @@
  * Raft session.
  */
 public class RaftSessionContext implements RaftSession {
+  private static final int PHI_FAILURE_THRESHOLD = 10;
```
kuujo: We should be able to remove this threshold constant

kuujo:
This implementation seems to have a problem with doing several rounds of election at startup. I haven't looked at the logs closely enough to figure out why yet.

kuujo:
nvm... I think it was because nodes weren't resetting election timers after voting in an election. Should be fixed now.

kuujo:
Rebased with #295

PR #296, opened 2017-11-04
Ensure expired session entries are applied on leader after commitment.

kuujo:
This PR fixes a bug in expiring sessions on the leader. The leader is not currently applying expire entries after they're committed. This prevents sessions from being expired until the next unrelated entry is applied, resulting in increased latency in expiring sessions.

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/roles/LeaderRole.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/roles/LeaderRole.java
@@ -281,7 +281,12 @@ public final class LeaderRole extends ActiveRole {
           }
 
           log.trace("Appended {}", entry);
-          appender.appendEntries(entry.index());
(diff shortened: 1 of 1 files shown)


protocols/raft/src/main/java/io/atomix/protocols/raft/roles/LeaderRole.java
```diff
@@ -281,7 +281,14 @@ private void expireSession(RaftSessionContext session) {
           }
 
           log.trace("Appended {}", entry);
-          appender.appendEntries(entry.index());
+          appender.appendEntries(entry.index()).whenComplete((commitIndex, commitError) -> {
+            raft.checkThread();
+            if (isOpen()) {
```
jhalterman: Can combine these two if statements :)

kuujo: Copying code always comes back to haunt me

PR #298, opened 2017-11-05
Update Netty messaging service to use new encoders/decoders

kuujo:


messaging/netty/src/main/java/io/atomix/messaging/netty/InternalMessage.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-present Open Networking Foundation
```
jhall11: Minor thing, but shouldn't this stay 2016-present?

PR #315, opened 2017-11-14
Support for map getAllPresent operation.

johnou:
In cases where one needs to check hundreds of entries are contained and / or what their values are in a distributed map iterating and collecting all futures is not optimal and can instead be done in a single operation (getAllPresent).


kuujo:
This is going to conflict with the PR I just merged

primitives/src/main/java/io/atomix/primitives/map/impl/RaftConsistentMapService.java
```diff
@@ -217,6 +220,19 @@ protected boolean containsValue(Commit<? extends ContainsValue> commit) {
...
 
   /**
+   * Handles a get all present commit.
+   *
+   * @param commit get all present commit
+   * @return keys present in map
+   */
+  protected Map<String, Versioned<byte[]>> getAllPresent(Commit<? extends GetAllPresent> commit) {
+    return entries().entrySet().stream()
+            .filter(entry -> commit.value().keys().contains(entry.getKey())
```
johnou: I'll flip the && here.

kuujo:
Code coverage tends to go up and down because of the non-deterministic nature of Raft.

PR #316, opened 2017-11-17
Fix log warning log format

y-higuchi:
Was leaving log entry like: 

 

> 2017-11-17 21:22:07,012 | WARN  | r-partition-1-18 | DefaultRaftServiceExecutor       | 90 - io.atomix - 2.0.6 | RaftService{645}{type=DOCUMENT_TREE-INSERTION, name=config-key-store} - State machine operation failed: {}


kuujo:
This needs to be pulled into 2.0.x

PR #319, opened 2017-12-01
Avoid repeated expiration of sessions on heartbeat failures

kuujo:
This PR ensures that leaders don't repeatedly attempt to expire the same session while waiting for an existing expiration change to be committed and applied. We do this by adding the `SessionId` to a `Set` prior to appending/replicating the `CloseSessionEntry` and only removing the `SessionId` once it has been committed/applied or failed to be committed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ notifications:
 branches:
   only:
     - master
+    - 2.0
 
 script:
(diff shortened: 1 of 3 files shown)


kuujo:
This needs to be cherry picked to `master` as well

PR #322, opened 2017-12-02
Do not register Raft message handlers until server becomes active

kuujo:
This PR resolves `IllegalStateException`s that can occur when clients/servers attempt to communicate with a node that's still in the `INACTIVE` state. We only register message handlers once the node has been started.


kuujo:
Actually, I think the `InactiveRole` should respond to requests with error responses.

PR #330, opened 2017-12-08
Ensure sessions are not expired prior to being registered

kuujo:
This PR fixes a bug in managing sessions in the Raft state machine. The problem is, when a session is first created it's visible to the primitive state machine's context before the session is open. In the event a command is executed on the state machine thread with a timestamp large enough to expire the session, the session may be expired before it's registered. This can lead to commands never being applied for the session. This bug primarily occurs when a node is replaying commands from the log, resulting in enough commands being applied to the state machine between the time the session is created and the time it's registered for it to be expired. The solution is to prevent unregistered sessions from being visible on the state machine thread.


kuujo:
Tested the last commit as well. I'm going to merge and release this for systems tests.

PR #333, opened 2017-12-14
Add ClusterMetadataService to track and replicate changes to DATA node membership

kuujo:
This PR adds a new `ClusterMetadataService` used to track and replicate the `DATA` node membership in the cluster. The service uses an anti-entropy protocol to replicate the membership and provides event listeners to react to changes in the set of data nodes.


cluster/src/main/java/io/atomix/cluster/impl/ClusterMetadataAdvertisement.java
```diff
@@ -0,0 +1,51 @@
...
+
+import io.atomix.cluster.NodeId;
+
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Cluster metadata anti-entropy advertisement.
+ */
+final class ClusterMetadataAdvertisement {
```
johnou: shouldn't this be serializable?

cluster/src/main/java/io/atomix/cluster/impl/ClusterMetadataAdvertisement.java
```diff
@@ -0,0 +1,51 @@
...
+    this.digests = digests;
+  }
+
+  /**
+   * Returns the set of digests in the advertisement.
+   *
+   * @return the set of digests in the advertisement
+   */
+  public Set<NodeId> digests() {
+    return digests.keySet();
```
johnou: be careful of this pitfall with kryo https://github.com/magro/kryo-serializers/issues/81

kuujo: That's interesting. And it feels familiar

cluster/src/main/java/io/atomix/cluster/impl/ClusterMetadataAdvertisement.java
kuujo: All the appropriate serializers needed for Kryo to handle the serialization are registered.

PR #334, opened 2017-12-15
Rename Atomix server to Atomix agent

kuujo:
The Atomix server is used as both a client and server, so perhaps the name is a bit of a misnomer. This PR refactors the server into an "agent" that supports both the client and server use cases.


electrical:
Failure is at 

```
[INFO] Running io.atomix.agent.AtomixAgentTest
[ERROR] Tests run: 3, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.703 s <<< FAILURE! - in io.atomix.agent.AtomixAgentTest
[ERROR] testParseNodeId(io.atomix.agent.AtomixAgentTest)  Time elapsed: 0.025 s  <<< FAILURE!
java.lang.AssertionError: expected:<localhost> but was:<ip6-localhost>
	at io.atomix.agent.AtomixAgentTest.testParseNodeId(AtomixAgentTest.java:43)
```

Which tends to happen when systems are IPV6 enabled. Depending on the hosts config file, often the ipv6 version of localhost is before the ipv4 one.
Perhaps a regex check if it contains `localhost` might be a bit easier and covering more systems.

johnou:
or perhaps something like this?

```java
assertEquals(NodeId.from(InetAddress.getLoopbackAddress().getHostName()), AtomixAgent.parseNodeId(new String[]{"127.0.0.1"}));
```
java.net.preferIPv4Stack system property is also an option but feels hacky.

electrical:
That might work as well yeah. Launching it with the IPv4 preference is indeed a bit hacky.

kuujo:
I'll try that thanks!

PR #335, opened 2017-12-15
Improve README documentation with various examples

kuujo:


README.md
```diff
@@ -5,140 +5,507 @@
...
+```java
+asyncElector.run("foo", atomix.getClusterService().getLocalNode().id()).thenAccept(leadership -> {
+  if (leadership.leader().id().equals(atomix.getClusterService().getLocalNode().id())) {
+    // Local node elected leader!
+  }
+});
+```
+
+The topic passed as the first argument to the `run` method is an identifier for the leader election
+to run into. The leader elector will balance the leaders for all topics.
```
jhall11: ... "method is an identifier for the leader election to run into". This seems confusing to me, maybe:

..." an identifier for the topic that the leader election runs for leadership of." ?

kuujo: Uhh... yeah I don't think that's what I meant to type :-P

PR #338, opened 2017-12-17
Clean up Atomix APIs

kuujo:
This PR cleans up various APIs within Atomix, changing the `Managed` interface method to `start()` and `stop()` rather than `open()` and `close()` and simplifying getters on the `Atomix` interface.


jplock:
Thanks for the quick response with this.  My sample project is working now after starting up 3 nodes in parallel (a quorum is reached and Jetty is able to start up).  I'd still like to request being able to start up a single node without specifying any bootstrap nodes though, as that would simplify things a bit.

PR #340, opened 2017-12-21
atomix-storage-parent removed

ChristerF:
It looks like the atomix-storage-parent pom was removed, repointing to the atomix-parent.


kuujo:
Probably need to clean out my .m2 directory :-P

Issue #344, opened 2017-12-22
Missing snapshot file exception


kuujo:
When recovering a Raft node from disk, infrequently the node may log the following `ERROR`:

```
2017-12-21 23:46:42,294 | ERROR | r-partition-5-20 | DefaultServiceContext            | 90 | RaftService{2}{type=CONSISTENT_MAP, name=onos-transactions} - Snapshot installation failed: {}
java.lang.IllegalStateException: missing snapshot file: /opt/onos/apache-karaf-3.0.8/data/db/partitions/5/onos-transactions-2-190940.snapshot
	at com.google.common.base.Preconditions.checkState(Preconditions.java:530)
	at io.atomix.protocols.raft.storage.snapshot.FileSnapshot.openReader(FileSnapshot.java:64)
	at io.atomix.protocols.raft.service.impl.DefaultServiceContext.maybeInstallSnapshot(DefaultServiceContext.java:257)
	at io.atomix.protocols.raft.service.impl.DefaultServiceContext.lambda$openSession$4(DefaultServiceContext.java:400)
	at io.atomix.utils.concurrent.ThreadPoolContext.lambda$new$0(ThreadPoolContext.java:81)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)[:1.8.0_101]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)[:1.8.0_101]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)[:1.8.0_101]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)[:1.8.0_101]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)[:1.8.0_101]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)[:1.8.0_101]
	at java.lang.Thread.run(Thread.java:745)[:1.8.0_101]
```

[...]

PR #348, opened 2017-12-24
Fix missing snapshot file exceptions

kuujo:
This PR fixes various inconsistencies that can occur in snapshots replicated from leaders to followers. When a leader replicates a snapshot to a follower, if that snapshot is at the same index as an existing snapshot on the follower, this can ultimately to the snapshot file being deleted. This is because old snapshots may be deleted once newer snapshots are persisted. But because both snapshots reference the same file, the persisting the new snapshot can ultimately mean deleting it. This PR also fixes a bug in snapshot replication that allows duplicate chunks sent by the leader. And finally, it disallows leaders from sending duplicate snapshots just to be safe.

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/service/impl/DefaultServiceContext.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/service/impl/DefaultServiceContext.java
@@ -41,6 +41,7 @@ import io.atomix.time.LogicalClock;
 import io.atomix.time.LogicalTimestamp;
 import io.atomix.time.WallClock;
 import io.atomix.time.WallClockTimestamp;
+import io.atomix.utils.concurrent.ComposableFuture;
 import io.atomix.utils.concurrent.ThreadContext;
 import io.atomix.utils.concurrent.ThreadContextFactory;
 import io.atomix.utils.logging.ContextualLoggerFactory;
@@ -237,7 +238,7 @@ public class DefaultServiceContext implements ServiceContext {
 
             // Update the snapshot index to ensure we don't simply install the same snapshot.
             snapshotIndex = snapshot.index();
-            pendingSnapshot.future.complete(null);
+            pendingSnapshot.future.complete(snapshotIndex);
           }
         }
       }
@@ -307,10 +308,11 @@ public class DefaultServiceContext implements ServiceContext {
    * @return a future to be completed once the snapshot has been taken
    */
   public CompletableFuture<Long> takeSnapshot(long index) {
-    CompletableFuture<Long> future = new CompletableFuture<>();
+    ComposableFuture<Long> future = new ComposableFuture<>();
     serviceExecutor.execute(() -> {
       // If no entries have been applied to the state machine, skip the snapshot.
       if (currentIndex == 0) {
+        future.complete(currentIndex);
         return;
       }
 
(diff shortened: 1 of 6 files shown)


kuujo:
Fixes #344

protocols/raft/src/main/java/io/atomix/protocols/raft/service/impl/DefaultServiceContext.java
```diff
@@ -360,6 +359,7 @@ private void maybeInstallSnapshot(long index) {
       // Persist the snapshot to disk in a background thread before completing the snapshot future.
       snapshotExecutor.execute(() -> {
         pendingSnapshot.persist();
+        pendingSnapshots.put(snapshotIndex, pendingSnapshot);
```
johnou: @kuujo why not just persist if return value of put is true?

johnou:
```java
  snapshotExecutor.execute(() -> {
    if (pendingSnapshots.put(snapshotIndex, pendingSnapshot)) {
      pendingSnapshot.persist();
    }
    future.complete(snapshotIndex);
  });
```

johnou:
```java
  snapshotExecutor.execute(() -> {
    if (pendingSnapshots.put(snapshotIndex, pendingSnapshot)) {
      pendingSnapshot.persist();
    }
    future.complete(snapshotIndex);
  });
```

kuujo: No need. This part of the logic actually works fine, and this particular change really has no effect other than to add `PendingSnapshot`s to the map a bit later so they're not unnecessarily checked in `maybeCompleteSnapshot`. Even if we only want to `persist()` the snapshot if there's not already a snapshot at the same index in the `pendingSnapshots` map because a snapshot could still exist at the index from a previous run or could have already been persisted and removed from the `pendingSnapshots` map. The primary fix in this PR is to handle duplicate snapshots replicated from the leader. But perhaps some additional checks should also be added to avoid duplicate snapshots being taken of the state machine as well.

johnou: The suggestion was mainly for aesthetic reasons, seems silly to invoke persist then add it to the pending map.

protocols/raft/src/main/java/io/atomix/protocols/raft/service/impl/DefaultServiceContext.java
```diff
@@ -317,23 +317,33 @@ private void maybeInstallSnapshot(long index) {
       // Compute the snapshot index as the greater of the compaction index and the last index applied to this service.
       long snapshotIndex = Math.max(index, currentIndex);
 
+      // If a snapshot is already persisted at this index, skip the snapshot.
+      Snapshot existingSnapshot = raft.getSnapshotStore().getSnapshot(serviceId, snapshotIndex);
+      if (existingSnapshot != null) {
+        return;
```
johnou: Shouldn't we be completing the future when returning early?

kuujo: Yep...

kuujo: Fixed

johnou:
lgtm

PR #349, opened 2017-12-24
Cherry pick snapshot changes into 2.1

kuujo:
Cherry pick new snapshot bug fixes into 2.1

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/service/RaftServiceContext.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/service/RaftServiceContext.java
@@ -320,21 +320,32 @@ public class RaftServiceContext implements ServiceContext {
       // Compute the snapshot index as the greater of the compaction index and the last index applied to this service.
       long snapshotIndex = Math.max(index, currentIndex);
 
+      // If a snapshot is already persisted at this index, skip the snapshot.
+      Snapshot existingSnapshot = raft.getSnapshotStore().getSnapshot(serviceId, snapshotIndex);
+      if (existingSnapshot != null) {
+        return;
+      }
+
       // If there's already a snapshot taken at a higher index, skip the snapshot.
       Snapshot currentSnapshot = raft.getSnapshotStore().getSnapshotById(primitiveId);
       if (currentSnapshot != null && currentSnapshot.index() >= index) {
         return;
       }
 
-      log.debug("Taking snapshot {}", snapshotIndex);
-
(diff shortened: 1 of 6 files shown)


kuujo:
#344

protocols/raft/src/main/java/io/atomix/protocols/raft/service/RaftServiceContext.java
```diff
@@ -310,33 +311,48 @@ private void maybeInstallSnapshot(long index) {
...
           .newTemporarySnapshot(primitiveId, serviceName, snapshotIndex, WallClockTimestamp.from(currentTimestamp));
 
-      // Add the snapshot to the pending snapshots registry.
+      // Add the snapshot to the pending snapshots registry. If a snapshot is already pending for this index,
+      // skip the snapshot.
       PendingSnapshot pendingSnapshot = new PendingSnapshot(snapshot);
-      pendingSnapshots.put(snapshotIndex, pendingSnapshot);
+      PendingSnapshot existingPendingSnapshot = pendingSnapshots.putIfAbsent(snapshotIndex, pendingSnapshot);
+      if (existingPendingSnapshot != null) {
+        existingPendingSnapshot.future.whenComplete(future);
```
johnou: Nice use of lambda exp.

protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/SnapshotStore.java
```diff
@@ -106,6 +109,22 @@ private void open() {
...
+   *
+   * @param serviceId the primitive for which to lookup the snapshot
+   * @param index     the index for which to lookup the snapshot
+   * @return the snapshot for the given service at the given index or {@code null} if the snapshot doesn't exist
+   */
+  public Snapshot getSnapshot(PrimitiveId serviceId, long index) {
+    Collection<Snapshot> snapshots = indexSnapshots.get(index);
+    return snapshots == null ? null :
+        snapshots.stream()
+            .filter(s -> s.serviceId().equals(serviceId))
```
johnou: sorted?

kuujo: The set ensures there's only one snapshot per service so `findFirst` or `findAny` is fine here. Sorting would just be unnecessary overhead. The reason the collection is sorted elsewhere is so the leader can get a consistent list of snapshots to iterate through at a given index when replicating snapshots since multiple snapshots for different state machines can exist at a single index.

PR #352, opened 2017-12-27
ThreadLocalRandom over Random.

johnou:
Random is thread safe for use by multiple threads. But if multiple threads use the same instance of Random, the same seed is shared by multiple threads. It leads to contention between multiple threads and so to performance degradation.

### DIFF
--- a/utils/src/main/java/io/atomix/utils/concurrent/Retries.java
+++ b/utils/src/main/java/io/atomix/utils/concurrent/Retries.java
@@ -15,7 +15,7 @@
  */
 package io.atomix.utils.concurrent;
 
-import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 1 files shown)


johnou:
Rebased.

kuujo:
I suspect these changes should not be necessary aside from the `Retries` change, which makes sense. The Raft roles have a single-threaded model that makes `Random` inherently thread-local, and we may just be introducing additional overhead by using a thread local object there.

johnou:
Sounds reasonable, updated PR as suggested.

utils/src/main/java/io/atomix/utils/concurrent/Retries.java
```diff
@@ -71,7 +70,7 @@
    */
   public static void randomDelay(int ms) {
     try {
-      Thread.sleep(random.nextInt(ms));
+      Thread.sleep(ThreadLocalRandom.current().nextInt(ms));
     } catch (InterruptedException e) {
       throw new RuntimeException("Interrupted", e);
```
johnou: Side node, we should be re-asserting the thread interrupted state.

PR #353, opened 2017-12-27
Correct handling of quorum timeout.

johnou:
if election times out the previous quorum was never cancelled..

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/roles/CandidateRole.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/roles/CandidateRole.java
@@ -42,7 +42,6 @@ import java.util.stream.Collectors;
  */
 public final class CandidateRole extends ActiveRole {
   private final Random random = new Random();
-  private Quorum quorum;
   private Scheduled currentTimer;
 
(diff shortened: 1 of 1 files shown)


protocols/raft/src/main/java/io/atomix/protocols/raft/roles/CandidateRole.java
```diff
@@ -124,6 +124,7 @@ private void sendVoteRequests() {
         raft.transition(RaftServer.Role.FOLLOWER);
       }
     });
+    this.quorum = quorum;
```
johnou: quorum was never set so if electrion times out the previous quorum was never cancelled.

johnou:
Don't merge yet, think I see a race condition.

protocols/raft/src/main/java/io/atomix/protocols/raft/roles/CandidateRole.java
johnou: @kuujo not sure if this is quite right, wdyt?

johnou:
@kuujo PTAL, hopefully this is what you had in mind.

kuujo:
Restarting tests

PR #355, opened 2017-12-27
Replace bad usage of SimpleDateFormat with threadsafe java 8 DateTime…

johnou:
…Formatter.

### DIFF
--- a/utils/src/main/java/io/atomix/utils/TimestampPrinter.java
+++ b/utils/src/main/java/io/atomix/utils/TimestampPrinter.java
@@ -15,8 +15,10 @@
  */
 package io.atomix.utils;
 
-import java.text.SimpleDateFormat;
-import java.util.Date;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
 
 /**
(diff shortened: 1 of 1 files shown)


utils/src/main/java/io/atomix/utils/TimestampPrinter.java
```diff
@@ -33,7 +35,7 @@ public static TimestampPrinter of(long timestamp) {
     return new TimestampPrinter(timestamp);
   }
 
-  private static final SimpleDateFormat FORMAT = new SimpleDateFormat("yyyy-MM-dd hh:mm:ss,SSS");
+  private static final DateTimeFormatter FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd hh:mm:ss,SSS");
```
johnou: instead of sync'ing on SimpleDateFormat or creating a new SimpleDateFormat or using threadlocal we can just use the new java.time API in java 8 which is thread safe.

johnou: Instead of sync'ing on the SimpleDateFormat instance, creating a new SimpleDateFormat or using threadlocal we can just use the new java.time API in java 8 which is thread safe.

PR #357, opened 2017-12-27
Fix bad builder.

johnou:


protocols/raft/src/main/java/io/atomix/protocols/raft/RaftServer.java
```diff
@@ -153,7 +154,7 @@
    * @return The server builder.
    */
   static Builder builder() {
-    return builder(null);
+    return builder(NodeId.from(InetAddress.getLoopbackAddress().getHostName()));
```
johnou: arg is passed directly into checkNotNull..

johnou: "localhost" might not be the best node id..

protocols/raft/src/main/java/io/atomix/protocols/raft/RaftServer.java
```diff
@@ -153,7 +156,12 @@
    * @return The server builder.
    */
   static Builder builder() {
-    return builder(null);
+    try {
+      InetAddress address = InetAddress.getByName("0.0.0.0");
```
johnou: consistent with io.atomix.core.Atomix.Builder#build

PR #358, opened 2017-12-27
sun.* is optional for utils artifact.

johnou:


utils/pom.xml
```diff
@@ -61,7 +61,7 @@
               io.atomix.utils.*
             </Export-Package>
             <Import-Package>
-              !sun.nio.ch,!sun.misc,*
+              sun.nio.ch;resolution:=optional,sun.misc;resolution:=optional,*
```
johnou: @kuujo I think this makes more sense as sun.* should only be required if using memory / mapped storage level, right?

PR #361, opened 2017-12-28
Fix equals methods.

johnou:


protocols/raft/src/main/java/io/atomix/protocols/raft/protocol/TransferRequest.java
```diff
@@ -58,10 +58,10 @@ public int hashCode() {
 
   @Override
   public boolean equals(Object object) {
-    if (getClass().isAssignableFrom(object.getClass())) {
-      return ((ConfigurationRequest) object).member.equals(member);
```
johnou: surprised this hasn't caused a class cast exception in the past.

PR #362, opened 2017-12-28
Release file input stream and use default type instead of hardcoded JKS.

johnou:


messaging/src/main/java/io/atomix/messaging/impl/NettyMessagingService.java
```diff
@@ -251,13 +251,17 @@ private boolean loadKeyStores() {
       char[] tsPwd = System.getProperty("javax.net.ssl.trustStorePassword", DEFAULT_KS_PASSWORD).toCharArray();
 
       tmf = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
-      KeyStore ts = KeyStore.getInstance("JKS");
-      ts.load(new FileInputStream(tsLocation), tsPwd);
+      KeyStore ts = KeyStore.getInstance(KeyStore.getDefaultType());
```
johnou: Returns the default keystore type as specified in the Java security properties file, or the string "jks" (acronym for "Java keystore") if no such property exists.

PR #363, opened 2017-12-28
Remove double string allocation and replace com.google.common.base.Ch…

johnou:
…arsets with StandardCharsets.

### DIFF
--- a/cluster/src/main/java/io/atomix/cluster/messaging/impl/ClusterMessage.java
+++ b/cluster/src/main/java/io/atomix/cluster/messaging/impl/ClusterMessage.java
@@ -15,12 +15,12 @@
  */
 package io.atomix.cluster.messaging.impl;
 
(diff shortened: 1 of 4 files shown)


cluster/src/main/java/io/atomix/cluster/messaging/impl/ClusterMessage.java
```diff
@@ -148,8 +148,8 @@ public static ClusterMessage fromBytes(byte[] bytes) {
     byte[] payloadBytes = new byte[buffer.getInt()];
     buffer.get(payloadBytes);
 
-    return new ClusterMessage(new NodeId(new String(senderBytes, Charsets.UTF_8)),
-        new String(new String(subjectBytes, Charsets.UTF_8)),
```
johnou: remanent of refactoring..?

PR #367, opened 2017-12-28
Remove unused fields, ensure channel is closed if logger fails, encod…

johnou:
…e string without invoking expensive getBytes.


messaging/src/main/java/io/atomix/messaging/impl/MessageDecoder.java
```diff
@@ -35,13 +35,13 @@
 
   private final Logger log = LoggerFactory.getLogger(getClass());
 
+  private static final byte[] EMPTY_PAYLOAD = new byte[0];
```
johnou: Safe as a primitive array with size zero is considered immutable :)

PR #372, opened 2018-01-02
Cluster reconfiguration tests

kuujo:
This PR adds new tests for reconfiguring the Atomix cluster as well as a few bug fixes for cluster reconfiguration.

First, the `ClusterMetadataService` does not properly handle single node clusters due to an issue in the `bootstrap` method. We simply check that there are nodes from which to bootstrap the configuration before bootstrapping it.

Second, a condition in the balancing of `RaftPartition`s on reconfiguration was not correct. This PR fixes that logic to ensure Raft nodes are added to/removed from partitions on metadata changes as necessary.

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/partition/RaftPartition.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/partition/RaftPartition.java
@@ -144,7 +144,7 @@ public class RaftPartition implements Partition<RaftProtocol> {
     if (server == null && metadata.members().contains(managementService.getClusterService().getLocalNode().id())) {
       server = createServer(managementService);
       return server.join(metadata.members());
-    } else if (server != null && server.isRunning()) {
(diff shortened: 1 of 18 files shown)


protocols/raft/src/main/java/io/atomix/protocols/raft/partition/RaftPartition.java
```diff
@@ -144,7 +144,7 @@ public File dataDirectory() {
     if (server == null && metadata.members().contains(managementService.getClusterService().getLocalNode().id())) {
       server = createServer(managementService);
       return server.join(metadata.members());
-    } else if (server != null && server.isRunning()) {
+    } else if (server != null && !metadata.members().contains(managementService.getClusterService().getLocalNode().id())) {
```
johnou: Arghh I was sniffing around this line

johnou:
Lgtm, nice clean up with the tests.

Issue #343, opened 2017-12-22
Cannot commit index error


kuujo:
When running cluster restart tests, I encountered what appears to be an inconsistency in the Raft log that prevents it from being read by a node:

```
2017-12-22 01:28:59,828 | ERROR | rver-partition-6 | RaftServiceManager               | 90 - io.atomix - 2.0.13.SNAPSHOT | RaftServer{partition-6} - Cannot commit index 54164
2017-12-22 01:29:02,578 | ERROR | rver-partition-6 | RaftServiceManager               | 90 - io.atomix - 2.0.13.SNAPSHOT | RaftServer{partition-6} - Cannot commit index 54165
2017-12-22 01:29:03,578 | ERROR | rver-partition-6 | RaftServiceManager               | 90 - io.atomix - 2.0.13.SNAPSHOT | RaftServer{partition-6} - Cannot commit index 54166
2017-12-22 01:29:04,078 | ERROR | rver-partition-6 | RaftServiceManager               | 90 - io.atomix - 2.0.13.SNAPSHOT | RaftServer{partition-6} - Cannot commit index 54167
```

A Raft node gets into this state and cannot apply any more changes. But upon restarting the node, the log recovered to normal.

kuujo:
This has been a really difficult issue to reproduce with any sort of consistency. I've only seen it that one time in testing and haven't been able to get good logs of the issue yet.

kuujo:
Reopening this issue. I've seen this again on a follower. The odd part about the pattern in which I see this occurring is that it _only_ happens on the last index committed by the follower. In other words, if the follower appends three entries `1`, `2`, and `3` and then sets the commit index to `3`, only index `3` will log the message `Cannot commit index 3`.

```
2018-01-03 20:35:15,037 | TRACE | rver-partition-2 | FollowerRole                     | 90 | RaftServer{partition-2}{role=FOLLOWER} - Received AppendRequest{term=4, leader=10.192.19.182, prevLogIndex=12178, prevLogTerm=4, entries=2, commitIndex=12178}
2018-01-03 20:35:15,037 | TRACE | rver-partition-2 | FollowerRole                     | 90 | RaftServer{partition-2}{role=FOLLOWER} - Appended Indexed{index=12179, entry=CommandEntry{term=4, timestamp=2018-01-03 08:35:15,034, session=8314, sequence=1418, operation=RaftOperation{id=UPDATE, value=byte[]{length=348, hash=1581998919}}}}
[...]

PR #373, opened 2018-01-03
Do not apply configuration change on replication failure

kuujo:
This PR fixes a missing error check that causes #343

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/roles/LeaderRole.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/roles/LeaderRole.java
@@ -341,7 +341,9 @@ public final class LeaderRole extends ActiveRole {
 
           return appender.appendEntries(entry.index()).whenComplete((commitIndex, commitError) -> {
             raft.checkThread();
-            raft.getStateMachine().<OperationResult>apply(entry.index());
+            if (isRunning() && commitError == null) {
+              raft.getStateMachine().<OperationResult>apply(entry.index());
+            }
             configuring = 0;
           });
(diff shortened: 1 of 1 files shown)


johnou:
Nice catch!

protocols/raft/src/main/java/io/atomix/protocols/raft/roles/LeaderRole.java
```diff
@@ -341,7 +341,9 @@ private boolean initializing() {
 
           return appender.appendEntries(entry.index()).whenComplete((commitIndex, commitError) -> {
             raft.checkThread();
-            raft.getStateMachine().<OperationResult>apply(entry.index());
+            if (isRunning() && commitError == null) {
```
johnou: Log warning? Debug?

kuujo: I've found this typically means the leader transitioned back to another role before the change could be committed, so other log messages already indicate what happened and the failure can just be ignored

kuujo: Hmm actually should probably still log it

johnou: At least debug.

kuujo: Actually, no logging is needed here. The _only_ case where this can fail is actually when a leader transitions back to another role, and the failure would result in an `ERROR` response to the reconfiguring node, so logging the error would be redundant.

PR #374, opened 2018-01-03
Default the number of Raft partitions to 7

kuujo:
This PR sets a default number of Raft partitions. This approach is being used here to simplify the configuration of Atomix clusters without introducing new complexity to the management of cluster metadata. We could compute and replicate the number of Raft partitions. But that will always be fraught with the same risks as manually configured clusters, and using a static default configuration allows for a practical number of partitions to be created for all clusters by default while still preserving the option to customize the number of Raft partitions. Because the primary-backup protocol uses a similar approach, this simplifies the configuration of clusters that will be modified by adding/removing nodes.

### DIFF
--- a/core/src/main/java/io/atomix/core/Atomix.java
+++ b/core/src/main/java/io/atomix/core/Atomix.java
@@ -286,7 +286,11 @@ public class Atomix implements PrimitivesService, Managed<Atomix> {
    */
   public static class Builder implements io.atomix.utils.Builder<Atomix> {
     protected static final String DEFAULT_CLUSTER_NAME = "atomix";
+    // Default to 7 Raft partitions to allow a leader per node in 7 node clusters
+    protected static final int DEFAULT_COORDINATION_PARTITIONS = 7;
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/atomix/core/Atomix.java
```diff
@@ -286,6 +286,7 @@ public String toString() {
    */
   public static class Builder implements io.atomix.utils.Builder<Atomix> {
     protected static final String DEFAULT_CLUSTER_NAME = "atomix";
+    protected static final int DEFAULT_COORDINATION_PARTITIONS = 7;
```
johnou: Why 7 and 71?

kuujo: 7 nodes is typical of large consensus based clusters and allows a leader per node, and 71 is a not-too-small, not-too-large prime number

johnou: code comment explaining the magic numbers?

Issue #375, opened 2018-01-03
Anti-entropy advertisements sent to tomestones


johnou:
Not sure if tomestone is the correct term, however, after a node leaves the cluster the remaining nodes continue to try and send anti-entropy advertisements to that node.
  
2018-01-03T11:18:35,530 WARN  [netty-messaging-event-nio-client-7] (io.atomix.cluster.impl.DefaultClusterMetadataService) Anti-entropy advertisement to 127.0.0.1:5003 failed! []
2018-01-03T11:18:36,530 WARN  [netty-messaging-event-nio-client-1] (io.atomix.cluster.impl.DefaultClusterMetadataService) Anti-entropy advertisement to 127.0.0.1:5003 failed! []
2018-01-03T11:18:37,443 WARN  [netty-messaging-event-nio-client-8] (io.atomix.cluster.impl.DefaultClusterMetadataService) Anti-entropy advertisement to 127.0.0.1:5003 failed! []
2018-01-03T11:18:38,531 WARN  [netty-messaging-event-nio-client-3] (io.atomix.cluster.impl.DefaultClusterMetadataService) Anti-entropy advertisement to 127.0.0.1:5003 failed! []
2018-01-03T11:18:39,443 WARN  [netty-messaging-event-nio-client-15] (io.atomix.cluster.impl.DefaultClusterMetadataService) Anti-entropy advertisement to 127.0.0.1:5003 failed! []
2018-01-03T11:18:41,531 WARN  [netty-messaging-event-nio-client-12] (io.atomix.cluster.impl.DefaultClusterMetadataService) Anti-entropy advertisement to 127.0.0.1:5003 failed! []
2018-01-03T11:18:42,444 WARN  [netty-messaging-event-nio-client-12] (io.atomix.cluster.impl.DefaultClusterMetadataService) Anti-entropy advertisement to 127.0.0.1:5003 failed! []
2018-01-03T11:18:42,531 WARN  [netty-messaging-event-nio-client-8] (io.atomix.cluster.impl.DefaultClusterMetadataService) Anti-entropy advertisement to 127.0.0.1:5003 failed! []
2018-01-03T11:18:44,532 WARN  [netty-messaging-event-nio-client-6] (io.atomix.cluster.impl.DefaultClusterMetadataService) Anti-entropy advertisement to 127.0.0.1:5003 failed! []
[...]

PR #376, opened 2018-01-03
Handle tomestones in cluster metadata service.

johnou:
Fixes https://github.com/atomix/atomix/issues/375

### DIFF
--- a/cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMetadataService.java
+++ b/cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMetadataService.java
@@ -257,8 +257,9 @@ public class DefaultClusterMetadataService
   private Optional<Endpoint> pickRandomPeer() {
     List<Endpoint> nodes = this.nodes.values()
         .stream()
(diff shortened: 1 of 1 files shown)


kuujo:
The problem here is the tombstones _need_ to be retained in the map so they can be replicated, otherwise other nodes will not learn about deleted nodes. If a node is simply removed from the map then that change will never be replicated. Tombstones need to be retained until some arbitrary point in the future to ensure deletes are replicated to all nodes. [This condition](https://github.com/johnou/atomix/blob/5831f7d2f7534a565d7d74bea62e80f034e432a5/cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMetadataService.java#L264-L265) is supposed to stop anti-entropy advertisements from being sent to removed nodes, and we need to figure out why that's not preventing anti-entropy advertisements from being sent to deleted nodes.

kuujo:
We should probably write a test for this and work back from there

johnou:
@kuujo couldn't we just revert everything except the changes to pickRandomPeer

johnou:
@kuujo PTAL

kuujo:
Yep that should work

PR #378, opened 2018-01-04
Implement support for time-to-live for map keys

kuujo:
This PR implements support for time-to-live durations for map keys, e.g.

```java
ConsistentMap<String, String> map = atomix.consistentMapBuilder("foo").build();
map.put("foo", "bar", Duration.ofSeconds(1));
Thread.sleep(2);
assert map.get("foo") == null;
```

### DIFF
--- a/core/src/main/java/io/atomix/core/map/AsyncConsistentMap.java
+++ b/core/src/main/java/io/atomix/core/map/AsyncConsistentMap.java
@@ -17,7 +17,6 @@
 package io.atomix.core.map;
 
 import com.google.common.util.concurrent.MoreExecutors;
-
 import io.atomix.core.PrimitiveTypes;
 import io.atomix.core.map.impl.MapUpdate;
 import io.atomix.core.transaction.Transactional;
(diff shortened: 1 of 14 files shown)


core/src/main/java/io/atomix/core/map/AsyncConsistentMap.java
```diff
@@ -213,7 +212,35 @@ default PrimitiveType primitiveType() {
    * @return the previous value (and version) associated with key, or null if there was
    * no mapping for key.
    */
-  CompletableFuture<Versioned<V>> put(K key, V value);
+  default CompletableFuture<Versioned<V>> put(K key, V value) {
+    return put(key, value, Duration.ZERO);
+  }
+
+  /**
+   * Associates the specified value with the specified key in this map (optional operation).
```
johnou: Is optional operation referring to the TTL?

core/src/main/java/io/atomix/core/map/impl/ConsistentMapService.java
```diff
@@ -342,6 +344,43 @@ protected boolean valueIsNull(MapEntryValue value) {
...
+
+  /**
+   * Schedules the TTL for the given value.
+   *
+   * @param value the value for which to schedule the TTL
+   */
+  protected void scheduleTtl(String key, MapEntryValue value) {
+    if (value.ttl() > 0) {
+      value.timer = scheduler().schedule(Duration.ofMillis(value.ttl() - value.created()), () -> {
+        entries().remove(key);
```
johnou: Wouldn't remove(key, value) be safer?

kuujo: Yeah good idea

core/src/main/java/io/atomix/core/map/AsyncConsistentMap.java
kuujo: `Duration.ZERO` creates no TTL

PR #380, opened 2018-01-06
Ensure clients are replicated to other clients in ClusterService

kuujo:
This PR fixes a deficiency in the `ClusterService` that did not properly replicate `CLIENT` type nodes to other `CLIENT` nodes. We use a pretty naive protocol to replicate client nodes, responding to heartbeats with all known clients. When a node learns about a client, it triggers node add/activate events. When a client fails, the client is retained as a tombstone but a remove event is triggered.

### DIFF
--- a/cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterService.java
+++ b/cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterService.java
@@ -36,6 +36,7 @@ import io.atomix.utils.serializer.KryoNamespaces;
 import io.atomix.utils.serializer.Serializer;
 import org.slf4j.Logger;
 
+import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
@@ -62,6 +63,7 @@ public class DefaultClusterService
 
   private static final int DEFAULT_HEARTBEAT_INTERVAL = 100;
   private static final int DEFAULT_PHI_FAILURE_THRESHOLD = 10;
+  private static final long DEFAULT_FAILURE_TIME = 1000;
   private static final String HEARTBEAT_MESSAGE = "atomix-cluster-heartbeat";
 
   private int heartbeatInterval = DEFAULT_HEARTBEAT_INTERVAL;
@@ -74,7 +76,10 @@ public class DefaultClusterService
           .nextId(KryoNamespaces.BEGIN_USER_CUSTOM_ID)
           .register(NodeId.class)
           .register(Node.Type.class)
+          .register(Node.State.class)
           .register(ClusterHeartbeat.class)
+          .register(StatefulNode.class)
+          .register(new DefaultClusterMetadataService.EndpointSerializer(), Endpoint.class)
           .build("ClusterService"));
 
   private final MessagingService messagingService;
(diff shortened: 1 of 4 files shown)


cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterService.java
```diff
@@ -104,12 +110,19 @@ public Node getLocalNode() {
...
+        .stream()
+        .filter(node -> node.type() == Node.Type.DATA || node.getState() == State.ACTIVE)
+        .collect(Collectors.toList()));
   }
 
   @Override
   public Node getNode(NodeId nodeId) {
-    return nodes.get(nodeId);
+    Node node = nodes.get(nodeId);
+    if (node != null && node.type() == Node.Type.CLIENT && node.getState() == State.INACTIVE) {
```
johnou: What about data type here?

johnou: If correct the method name should be clearer.

cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterService.java
```diff
@@ -154,10 +181,13 @@ private void sendHeartbeat(Endpoint endpoint, byte[] payload) {
   /**
    * Handles a heartbeat message.
    */
-  private void handleHeartbeat(Endpoint endpoint, byte[] message) {
+  private byte[] handleHeartbeat(Endpoint endpoint, byte[] message) {
     ClusterHeartbeat heartbeat = SERIALIZER.decode(message);
     failureDetectors.computeIfAbsent(heartbeat.nodeId(), n -> new PhiAccrualFailureDetector()).report();
     activateNode(new StatefulNode(heartbeat.nodeId(), heartbeat.nodeType(), endpoint));
+    return SERIALIZER.encode(nodes.values().stream()
+        .filter(node -> node.type() == Node.Type.CLIENT)
```
johnou: Data type?

core/src/test/java/io/atomix/core/AtomixTest.java
```diff
@@ -82,4 +89,80 @@ public void testScaleDown() throws Exception {
...
+  private static class TestClusterEventListener implements ClusterEventListener {
+
+    private final BlockingQueue<ClusterEvent> queue = new ArrayBlockingQueue<>(1);
+
+    @Override
+    public void onEvent(ClusterEvent event) {
+      try {
+        queue.put(event);
+      } catch (InterruptedException e) {
+        Throwables.propagate(e);
```
johnou: Why propagate and not simply Thread.currentThread().interrupt()

cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterService.java
kuujo: This logic is just filtering tombstones, which shouldn't be visible via the API. It's an implementation detail of the replication algorithm.

core/src/test/java/io/atomix/core/AtomixTest.java
kuujo: Just catching a checked exception and rethrowing it as an unchecked exception

cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterService.java
kuujo: Data nodes are replicated by the `ClusterMetadataService` and would be redundant, so only client nodes are filtered and sent back in response to heartbeats

core/src/test/java/io/atomix/core/AtomixTest.java
johnou: I think we should restore the interrupted status

johnou: https://www.ibm.com/developerworks/library/j-jtp05236/index.html

PR #381, opened 2018-01-06
Fix improper lock override on tryLock()

kuujo:
This PR fixes bugs in the `DistributedLock` primitive state machine. When `tryLock` is used, an existing lock can be improperly overwritten by the new lock.

### DIFF
--- a/core/src/test/java/io/atomix/core/lock/impl/DistributedLockTest.java
+++ b/core/src/test/java/io/atomix/core/lock/impl/DistributedLockTest.java
@@ -21,9 +21,11 @@ import io.atomix.utils.time.Version;
 import org.junit.Test;
 
 import java.time.Duration;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 
 import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/atomix/core/lock/impl/DistributedLockTest.java
```diff
@@ -63,9 +65,46 @@ public void testTryLockFail() throws Throwable {
...
+
+    lock1.lock().join();
+
+    CompletableFuture<Optional<Version>> future = lock2.tryLock(Duration.ofSeconds(1));
+    lock1.unlock().join();
+    assertTrue(future.join().isPresent());
+  }
+
   /**
    * Tests unlocking a lock with a blocking call in the event thread.
```
johnou: Should we also assert is present for the joins in testBlockingUnlock?

kuujo: yep

johnou: Reviewing from mobile ftl :d

PR #384, opened 2018-01-08
Ensure map TTLs are correctly computed when initializing and restoring from snapshots

### LINKED ISSUE #382: ConsistentMapp TTL does not work as expected
If a TTL is set, the value can be retrieved immediately, but if some time (< TTL time) passes, the get could return null.

Reproducer: Set a timeout of 3 seconds, put a value in, sleep for 1 second, value will be gone

Replace the following on line 227 of  _ConsistentMapTest.java_

````
map.put("foo", "Hello foo!", Duration.ofSeconds(1)).thenAccept(result -> {
  assertNull(result);
}).join();
````
With 
````
map.put("foo", "Hello foo!", Duration.ofSeconds(3)).thenAccept(result -> {
  assertNull(result);
}).join();

Thread.sleep(1000);
````
This will fail the test, but the value should not have been removed yet after 1 second, but 3

### PR DESCRIPTION
**Pull request by kuujo**
Fixes #382 

The problem was that there was an odd computation being used to schedule the value expiration. I suspect the reads in tests were just being performed before the next tick of the scheduler. Additionally, this PR ensures the elapsed time is properly calculated into the TTL when restoring from a snapshot.

### DIFF
--- a/core/src/main/java/io/atomix/core/map/impl/ConsistentMapService.java
+++ b/core/src/main/java/io/atomix/core/map/impl/ConsistentMapService.java
@@ -153,7 +153,14 @@ public class ConsistentMapService extends AbstractPrimitiveService {
     map = reader.readObject(serializer()::decode);
     activeTransactions = reader.readObject(serializer()::decode);
     currentVersion = reader.readLong();
-    map.forEach(this::scheduleTtl);
+    map.forEach((key, value) -> {
+      if (value.ttl() > 0) {
+        value.timer = scheduler().schedule(Duration.ofMillis(value.ttl() - (wallClock().time().unixTimestamp() - value.created())), () -> {
+          entries().remove(key, value);
+          publish(new MapEvent<>(MapEvent.Type.REMOVE, "", key, null, toVersioned(value)));
+        });
+      }
+    });
   }
 
   @Override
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: core/src/main/java/io/atomix/core/map/impl/ConsistentMapService.java [hunk 1]
```diff
@@ -153,7 +153,14 @@ public void restore(BufferInput<?> reader) {
     map = reader.readObject(serializer()::decode);
     activeTransactions = reader.readObject(serializer()::decode);
     currentVersion = reader.readLong();
-    map.forEach(this::scheduleTtl);
+    map.forEach((key, value) -> {
+      if (value.ttl() > 0) {
+        value.timer = scheduler().schedule(Duration.ofMillis(value.ttl() - (wallClock().time().unixTimestamp() - value.created())), () -> {
```
johnou: Do we protect against negative milliseconds?

kuujo: I thought about that when I wrote this: the correctness of Raft and the state machine should guarantee that a negative duration is impossible. For that reason, we should rely on a negative TTL causing an exception, indicating an inconsistency in Raft or the state machine implementation.

On a related node, the scheduler needs the appropriate negative duration checks.

kuujo: #385

johnou: Ok so the correctness of Raft and the state machine should avoid the situation of trying to schedule expiration of an expired entry on restoration..?

kuujo: Correct. A snapshot should always be taken and installed at the same point in logical time, therefore it should not be possible for an entry that should have expired to be present in the snapshot. The `wallClock()` is actually a deterministic approximation of wall clock time that’s based on times in the Raft log.

PR #385, opened 2018-01-08
Add assertions to prevent negative TTLs in default service scheduler

kuujo:
This PR adds assertions needed to prevent negative `Duration`s in deterministic service schedulers. See #384


johnou:
Nice clean up.

kuujo:
Tests are failing because of negative TTLs :-)

PR #387, opened 2018-01-09
Use enum names for all primitive operation IDs

kuujo:
This PR changes all primitive operations to use the enum `name()` for consistency with [ONOS](http://onosproject.org) primitives. This is needed by ONOS to ensure the on-disk representation of Raft operations does not change across versions for rolling upgrades.

### DIFF
--- a/core/src/main/java/io/atomix/core/counter/impl/AtomicCounterOperations.java
+++ b/core/src/main/java/io/atomix/core/counter/impl/AtomicCounterOperations.java
@@ -21,30 +21,31 @@ import io.atomix.utils.serializer.KryoNamespace;
 import io.atomix.utils.serializer.KryoNamespaces;
 
 /**
- * Counter commands.
+ * {@link io.atomix.core.counter.AtomicCounter} operations.
+ * <p>
+ * WARNING: Do not refactor enum values. Only add to them.
+ * Changing values risk breaking the ability to backup/restore/upgrade clusters.
  */
 public enum AtomicCounterOperations implements OperationId {
-  SET("set", OperationType.COMMAND),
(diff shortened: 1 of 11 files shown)


kuujo:
Fair point. But really, the original implementation had the same problem, which is why I changed it in ONOS in the first place. Some of the `String` names were actually not accurately representative of the command names. But refactoring the string constant also changes the serialization. At least this ensures the serialized names are representative of the actual commands they execute, although it still suffers from the refactoring issue...

Documentation incoming

johnou:
The other option is of course changing the id from string to an int or even short and using that for serialisation as enum ordinal suffers from additional issues eg. ordering.

kuujo:
We used to use numeric IDs, but that makes the cluster very difficult to debug, e.g.

```
Operation '3' failed
```

kuujo:
Also it's too late. This can't be changed for now. We've released ONOS with `String` operation IDs and need to preserve the ability to perform rolling upgrades with the new version of Atomix and would prefer not to have to write scripts to process logs.

kuujo:
We've also already released a version of ONOS with string operation IDs and need to preserve the ability to do rolling upgrades, preferably without needing to enable compatible serialization just to handle operation names. This PR is just making Atomix 2.1 consistent with logs already stored in ONOS to ensure its primitives are compatible with equivalent Atomix primitives for upgrades. We can revisit modifying the structure of state stored on disk once all refactoring is done and there's less need to have rich debugging information in the logs (i.e. operation names)

johnou:
Making it compatible makes sense. I would still argue persisting a compact data type would be beneficial, logging could be improved to log the enum name from id. I agree revisiting the data structure later if necessary.

Issue #388, opened 2018-01-09
Use primitive exceptions in Raft implementation


kuujo:
The Raft implementation still uses exceptions and error codes that are separate from the primitive framework. This leads to some incorrect exceptions being handled and is causing `UnknownSession` errors in tests. One set of primitive exceptions needs to be used for all primitive protocols.

PR #389, opened 2018-01-09
Transform Raft exceptions into PrimitiveException on client

kuujo:
This PR transforms Raft exceptions into `PrimitiveException` on clients to ensure `RetryingPrimitiveProxy` properly handles exceptions.


kuujo:
#388

kuujo:
`RaftException` is internal to the Raft implementation as is used far too many places to be removed. We have to be really careful to ensure error handling is not changed.

johnou:
sgtm

PR #391, opened 2018-01-09
Use ClusterService for leader/session failure detection in RaftServer

kuujo:
This PR modifies the `RaftServer` to use the `ClusterService` for failure detection. The Raft implementation in the past has used a separate `PhiAccrualFailureDetector` to detect client failures and expire sessions. However, since it now has access to the `ClusterService` and its general failure detection, we should use the existing failure detector rather than unnecessarily adding additional network overhead for Raft failure detection.

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/RaftServer.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/RaftServer.java
@@ -15,6 +15,7 @@
  */
 package io.atomix.protocols.raft;
 
+import io.atomix.cluster.ClusterService;
 import io.atomix.cluster.NodeId;
 import io.atomix.primitive.PrimitiveType;
 import io.atomix.primitive.PrimitiveTypeRegistry;
@@ -41,7 +42,7 @@ import java.util.function.Supplier;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
-import static io.atomix.protocols.raft.RaftException.*;
+import static io.atomix.protocols.raft.RaftException.ConfigurationException;
 
 /**
  * Provides a standalone implementation of the <a href="http://raft.github.io/">Raft consensus algorithm</a>.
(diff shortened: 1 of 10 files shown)


protocols/raft/src/main/java/io/atomix/protocols/raft/RaftServer.java
```diff
@@ -598,6 +596,17 @@ public Builder withName(String name) {
...
     }
 
+    /**
+     * Sets the cluster service.
+     *
+     * @param clusterService the cluster service
+     * @return the server builder
+     */
+    public Builder withClusterService(ClusterService clusterService) {
+      this.clusterService = checkNotNull(clusterService, "clusterService cannot be null");
```
johnou: Verify non null in the builder when building?

kuujo: It's done in the `RaftContext` constructor

protocols/raft/src/main/java/io/atomix/protocols/raft/roles/LeaderRole.java
```diff
@@ -190,112 +185,28 @@ private void appendMembers() {
...
-        });
-      }
-    }, raft.getThreadContext());
-  }
-
   /**
    * Expires the given session.
    */
   private void expireSession(RaftSession session) {
     log.debug("Expiring session due to heartbeat failure: {}", session);
```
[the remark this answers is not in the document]
kuujo: did

protocols/raft/src/main/java/io/atomix/protocols/raft/roles/FollowerRole.java
```diff
@@ -134,7 +130,6 @@ private void sendPollRequests() {
       // If a majority of the cluster indicated they would vote for us then transition to candidate.
       complete.set(true);
       if (elected) {
-        raft.setLeader(null);
```
johnou: @kuujo handled elsewhere now?

Issue #366, opened 2017-12-28
φ (phi) timeout false positive


johnou:
When running a simple stress test I received a timeout of 8 milliseconds which is far too strict.

```
Caused by: java.util.concurrent.TimeoutException: Request timed out in 8 milliseconds
at io.atomix.messaging.impl.NettyMessagingService$RemoteClientConnection.timeoutCallbacks(NettyMessagingService.java:879)
```

johnou:
Maybe response times < 100ms should be excluded from samples..?

johnou:
Option for disabling PHI timeout might be nice too.

johnou:
Akka implementation uses "minStdDeviation" [1] which might help in this case.

```
 * @param minStdDeviation Minimum standard deviation to use for the normal distribution used when calculating phi.
 *   Too low standard deviation might result in too much sensitivity for sudden, but normal, deviations
 *   in heartbeat inter arrival times.
```

[1]
https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/PhiAccrualFailureDetector.scala#L182

kuujo:
That seems worth a test.

johnou:
I'll check it out tomorrow 👍

johnou:
Excluding reply time of zero from samples was enough for my stress test, didn't need to change the PHI calculation logic eg. factor in CV.

johnou:
https://github.com/atomix/atomix/pull/400 is good for now.

PR #398, opened 2018-01-16
Do not add reply time of zero to samples.

johnou:
By excluding reply time of zero in samples the custom PHI failure detector in NettyMessagingService did not report any false positives with my previous stress test and test with artificial random delays.

Fixes https://github.com/atomix/atomix/issues/366


johnou:
Closing in favour of https://github.com/atomix/atomix/pull/400

PR #400, opened 2018-01-16
Add minimum timeout for Netty messages

kuujo:
Adds a minimum timeout of 100ms for Netty messages as a temporary solution to #366


johnou:
Lgtm

PR #404, opened 2018-01-22
Netty 4.1.21.Final

johnou:
Contains fix for out of order messaging.

https://github.com/netty/netty/pull/7563

### DIFF
--- a/messaging/pom.xml
+++ b/messaging/pom.xml
@@ -35,22 +35,19 @@
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-transport</artifactId>
-      <version>${netty.version}</version>
     </dependency>
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-codec</artifactId>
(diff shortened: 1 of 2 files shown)


johnou:
Depending on usage this version may cause the jvm to hang on shutdown.. https://github.com/netty/netty/pull/7627

messaging/pom.xml
```diff
@@ -35,22 +35,19 @@
...
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-handler</artifactId>
-      <version>${netty.version}</version>
     </dependency>
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-transport-native-epoll</artifactId>
-      <version>${netty.version}</version>
+      <classifier>linux-x86_64</classifier>
```
johnou: @kuujo the artifact without the classifier doesn't actually contain any native transports, epoll should now work on Linux x64 out of the box.

Issue #405, opened 2018-01-24
Realization of φ accrual failure detector


baymaxhuang:
The [PhiAccrualFailureDetector](https://github.com/atomix/atomix/blob/832df3702d1aabbbbb1a2e3e395d50b1d0e94972/cluster/src/main/java/io/atomix/cluster/impl/PhiAccrualFailureDetector.java) in Atomix is actually implemented based on the [FailureDetector](https://github.com/apache/cassandra/blob/9b828275288b135812a247c3f2e9da2f20547111/src/java/org/apache/cassandra/gms/FailureDetector.java) of Cassandra, which is a little different from the original Phi failure detector proposed in the original paper by Hayashibara et al. See: https://issues.apache.org/jira/browse/CASSANDRA-2597

johnou:
@baymaxhuang so what's the actual problem? the implementations look very different to me, what makes you think that the Atomix implementation was based on the Cassandra code?

baymaxhuang:
The original phi value in Hayashibara's paper is calculated based on the normal distribution, the failure detector in [Akka](https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/PhiAccrualFailureDetector.scala#L173-L189) uses the original realization. However, the phi value in Atomix is calculated based on an exponential distribution, which is the same as the realization in Cassandra. The [Cassandra paper](http://www.cs.cornell.edu/projects/ladis2009/papers/lakshman-ladis2009.pdf) has mentioned why a exponential distribution is applied and there is also a discussion in [CASSANDRA-2597](https://issues.apache.org/jira/browse/CASSANDRA-2597).

johnou:
Are you suggesting the implementation could be improved? Pull requests are welcome!

baymaxhuang:
I think the code comment should be improved to make it more clear. The realization could remain what it was. I will pull a request.

johnou:
@baymaxhuang good to close this issue?

PR #406, opened 2018-01-24
Make the comments of phi failure detector clearer

baymaxhuang:
See [issue 405](https://github.com/atomix/atomix/issues/405) for more details.

### DIFF
--- a/cluster/src/main/java/io/atomix/cluster/impl/PhiAccrualFailureDetector.java
+++ b/cluster/src/main/java/io/atomix/cluster/impl/PhiAccrualFailureDetector.java
@@ -22,7 +22,8 @@ import static com.google.common.base.Preconditions.checkArgument;
 /**
  * Phi Accrual failure detector.
  * <p>
- * Based on a paper titled: "The φ Accrual Failure Detector" by Hayashibara, et al.
+ * A modified version based on a paper titled:
+ * "The φ Accrual Failure Detector" by Hayashibara, et al.
  */
 public class PhiAccrualFailureDetector {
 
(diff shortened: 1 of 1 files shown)


cluster/src/main/java/io/atomix/cluster/impl/PhiAccrualFailureDetector.java
```diff
@@ -22,7 +22,9 @@
 /**
  * Phi Accrual failure detector.
  * <p>
- * Based on a paper titled: "The φ Accrual Failure Detector" by Hayashibara, et al.
+ * A modified version based on a paper titled:
+ * "The φ Accrual Failure Detector" by Hayashibara, et al.
+ * See <a href="https://github.com/atomix/atomix/issues/405">issue 405</a> for more details.
```
johnou: I don't think linking to the Github issue is necessary.

cluster/src/main/java/io/atomix/cluster/impl/PhiAccrualFailureDetector.java
```diff
@@ -126,6 +128,7 @@ public double phi() {
    * @param currentTime   the current time
    * @return phi
    */
+  // In this version, phi value is calculated based on an exponential distribution.
```
johnou: should be part of the javadoc.

cluster/src/main/java/io/atomix/cluster/impl/PhiAccrualFailureDetector.java
kuujo: Concur

johnou:
@baymaxhuang ping

baymaxhuang:
@coveralls I have updated it with a new patch set.

PR #408, opened 2018-01-29
Check for closeFuture when start is invoked.

johnou:
Simple check to avoid confusion when invoking start after stop (not supported).


johnou:
Right now there are multiple services that shutdown executors which I think are fine to be left as is, I don't see any use-cases for reusing an Atomix instance, you can always recreate it with the same settings.

PR #410, opened 2018-02-03
Extra debug logging on start.

johnou:
Helpful to see where the node is up to in case of failure when starting and how long each service takes.


kuujo:
Closing this since we should have `Started` Messages logged by each indiviual service’s logger. We’ll have to revisit/improve logging in general soon though.

PR #413, opened 2018-02-05
If localNode is not specified bind to first NIC and if that fails to …

johnou:
…localhost.


johnou:
Same behaviour as JGroups if the bind address is not set.

PR #415, opened 2018-02-06
Support for BiConsumer handler which accepts endpoint and payload + reduce allocating strings.

johnou:


cluster/src/main/java/io/atomix/cluster/messaging/impl/ClusterMessage.java
```diff
@@ -153,6 +153,17 @@ public static ClusterMessage fromBytes(byte[] bytes) {
         payloadBytes);
   }
 
+  static byte[] getPayload(byte[] bytes) {
```
johnou: stripped down version of fromBytes which doesn't allocate char arrays as they were not used in the apply callbacks.

PR #416, opened 2018-02-06
Reduce copies and allocating strings.

johnou:


johnou:
@slandelle could I trouble you for a review?

johnou:
@slandelle makes sense, so something like this? (keeping it simple without supporting composite buffers)

```java
  private static byte[] readBytes(ByteBuf byteBuf, int length) {
    if (byteBuf.hasArray()) {
      final int offset = byteBuf.arrayOffset() + byteBuf.readerIndex();
      final byte[] bytes;
      if (offset == 0 && length == byteBuf.array().length) {
        bytes = byteBuf.array();
      } else {
        bytes = Arrays.copyOfRange(byteBuf.array(), offset, offset + length);
      }
      byteBuf.skipBytes(length);
      return bytes;
    }
    byte[] array = new byte[length];
    byteBuf.readBytes(array);
    return array;
  }
```

slandelle:
@johnou Why do you want to get bytes while subject is an UF-8 string?

You can directly create the String from the bytes: `new String(byteBuf.array(), offset, length)`

johnou:
@slandelle oh man I forgot that constructor, ptal.

messaging/src/main/java/io/atomix/messaging/impl/MessageDecoder.java
```diff
@@ -148,6 +185,29 @@ protected void decode(
...
+  static String readString(ByteBuf buffer, int length, Charset charset) {
+    if (buffer.isDirect()) {
+      final String result = buffer.toString(buffer.readerIndex(), length, charset);
+      buffer.skipBytes(length);
+      return result;
+    } else {
+      if (buffer.hasArray()) {
+        final int offset = buffer.arrayOffset() + buffer.readerIndex();
+        final String result;
+        if (offset == 0 && length == buffer.array().length) {
```
slandelle: IMHO, testing this is useless and you should just perform the second branch.

messaging/src/main/java/io/atomix/messaging/impl/MessageDecoder.java
```diff
@@ -148,6 +185,29 @@ protected void decode(
...
+          result = new String(buffer.array(), charset);
+        } else {
+          result = new String(buffer.array(), offset, length, charset);
+        }
+        buffer.skipBytes(length);
+        return result;
+      }
+      final byte[] array = new byte[length];
```
slandelle: Complicated branches. Why not if(isDirect) elseif (hasArray) else?

messaging/src/main/java/io/atomix/messaging/impl/MessageDecoder.java
johnou: removed.

messaging/src/main/java/io/atomix/messaging/impl/MessageDecoder.java
johnou: sgtm, updated.

johnou:
@slandelle thanks for the review!

kuujo:
Why are we replacing `ReplayingDecoder` with custom checkpointing?

johnou:
http://normanmaurer.me/presentations/2014-facebook-eng-netty/slides.html#13.0

More overhead and requires handling ReplayingError.

kuujo:
Nice sounds good

PR #418, opened 2018-02-06
Numerous fault tolerance bug fixes

kuujo:
• A race condition resulting from the order in which snapshots are applied to state machines and the ordering of the snapshot replication protocol can lead to a replicated snapshot being skipped rather than installed. This sometimes results in Raft sessions not being properly restored after a node is down for a long time before being restarted. Nodes now ensure they apply the latest snapshot before applying the following index rather than applying the next snapshot after the prior index to ensure snapshots replicated between the two indexes are installed when the next index is applied.
• Servers send ordering information to clients with each event, and clients use that information to ensure events and responses are handled in the correct order. But currently, servers send the last completed event index to clients with server events rather than the previous event index if the last completed index exceeds the previous event index. Because snapshots are always completed after all events are received, this logic can lead servers to send a non-event index to clients after a snapshot is installed and thus cause clients to wait for an event that will never arrive, ultimately resulting in a livelock and timeouts. The event protocol now always uses the previous event index when sending events to clients.
• After snapshots are taken of the system’s state and early segments are removed from the log, log readers’ indexes may not be properly reset to the head of the log. This can lead to leaders’ log readers for unavailable followers attempting to read closed/deleted segments once those followers recover, and prevents log segments indexes from aligning properly. Log readers’ heads are now properly reset after compaction.
• After a leader replicates a set of snapshots to a follower, the last replicated snapshot ID is not correctly reset. This prevents a leader from replicating more than one set of snapshots to the same follower. Leaders now properly reset snapshot IDs after the snapshot replication protocol is complete.
• Leaders currently allow the matchIndex for a follower to be incorrectly *increased* on a failed AppendResponse. The rationale for modifying the matchIndex on a failed AppendResponse is to account for missing entries after restarting a follower that uses MEMORY logs or doesn’t flush commits to disk. But matchIndex should only be allowed to be *decreased* by a failed AppendResponse.
• When a follower is recovering from a large log (e.g. hundreds of thousands of commands), it can still begin an election and get elected leader. This can lead to blocked commands while the log was replayed, causing timeouts on clients. Followers no longer attempt to get elected while sill replaying logs.
• When a follower is recovering from a large log (e.g. hundreds of thousands of commands), it accepts and queues SEQUENTIAL queries. This can also lead to query timeouts while logs were being replayed. Followers no longer accept queries while logs are being replayed.
• When a leader is pipelining entries to a follower with an inconsistent log (that needs to be truncated), failed AppendResponses in some cases do not cause the leader to correctly reset the follower’s log reader before the next request because of the order in which the leader handles the response and sends the next AppendRequest. If linearizable queries are pending on the leader, it may send the next AppendRequest before handling the response and resetting the follower’s log reader. Pipelining of AppendRequests can then cause a loop between the leader and follower wherein the follower’s log is never properly resolved with the leader’s, ultimately leading to an additional leader change and timeouts on clients. Leaders will now always handle responses before attempting to send any further entries to ensure log readers are properly managed based on response context.
• The expensive cost of resetting log readers’ indexes (seeking through potentially large log segments) on followers after a new leader is elected can lead to AppendRequest timeouts when large log segments are used. Log readers’ indexes are now checked before being reset on followers to avoid resetting a log reader to the same index.
• While replaying log entries at startup, a node’s request thread can be blocked, leading to leader election timeouts and unnecessary leader changes. Log entry application is now done asynchronously via a queue to avoid blocking.

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
@@ -124,8 +124,8 @@ public class RaftServiceManager implements AutoCloseable {
       if (nextIndex < index) {
         Indexed<RaftLogEntry> entry = reader.next();
         try {
+          restoreIndex(entry.index() - 1);
           apply(entry);
-          restoreIndex(entry.index());
         } catch (Exception e) {
           logger.error("Failed to apply {}: {}", entry, e);
         } finally {
@@ -141,9 +141,8 @@ public class RaftServiceManager implements AutoCloseable {
           if (entry.index() != index) {
             throw new IllegalStateException("inconsistent index applying entry " + index + ": " + entry);
           }
-          CompletableFuture<T> future = apply(entry);
-          restoreIndex(entry.index());
-          return future;
+          restoreIndex(entry.index() - 1);
+          return apply(entry);
         } catch (Exception e) {
           logger.error("Failed to apply {}: {}", entry, e);
         } finally {
@@ -152,6 +151,7 @@ public class RaftServiceManager implements AutoCloseable {
       }
       // If the applied index has been passed, return a null result.
       else {
+        logger.warn("Skipped applying index {}", index);
         raft.setLastApplied(nextIndex);
         return Futures.completedFuture(null);
       }
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
@@ -15,6 +15,7 @@
  */
 package io.atomix.protocols.raft.impl;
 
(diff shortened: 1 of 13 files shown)


kuujo:
Ropened against the correct branch

protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
```diff
@@ -69,6 +71,7 @@
   private final ThreadContextFactory threadContextFactory;
   private final RaftLog log;
   private final RaftLogReader reader;
+  private final Map<Long, CompletableFuture> futures = Maps.newHashMap();
```
johnou: @kuujo shouldn't this be a concurrent hash map?

kuujo: The state machine is single threaded since changes have to be applied in order

PR #419, opened 2018-02-06
Ensure sequential queries account for received events

kuujo:
`SEQUENTIAL` queries from Raft clients currently break the sequential consistency model with regard to events. When a client receives an event at index `i`, its next query must occur at index `>= i`. This PR simply uses `max(highestResponseIndex, highestEventIndex)` when sending queries to the cluster.


johnou:
Lgtm

johnou:
@kuujo do we need to revert this in 2.0 and master?

kuujo:
No, that would just break consistency again. This is an important consistency fix. All this does is ensures clients’ reads occur after the last event received so e.g. when a client receives an event saying a key was added to a map, the key can be immediately read by the client.

PR #422, opened 2018-02-10
Ensure messages are correctly timed out in NettyMessagingManager.

johnou:
Ensure Netty channels are closed on send exceptions.

Took updates from ONOS and discarded logger changes.

### DIFF
--- a/messaging/src/main/java/io/atomix/messaging/impl/NettyMessagingService.java
+++ b/messaging/src/main/java/io/atomix/messaging/impl/NettyMessagingService.java
@@ -15,6 +15,7 @@
  */
 package io.atomix.messaging.impl;
 
+import com.google.common.base.Throwables;
 import com.google.common.cache.Cache;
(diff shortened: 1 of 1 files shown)


messaging/src/main/java/io/atomix/messaging/impl/NettyMessagingService.java
```diff
@@ -999,10 +1063,9 @@ private void dispatch(InternalRequest message) {
 
       BiConsumer<InternalRequest, ServerConnection> handler = handlers.get(message.subject());
       if (handler != null) {
-        log.trace("{} - Received message type {} from {}", localEndpoint, message.subject(), message.sender());
         handler.accept(message, this);
       } else {
-        log.debug("{} - No handler for message type {} from {}", localEndpoint, message.subject(), message.sender());
```
johnou: should we revert the logging changes?

PR #423, opened 2018-02-11
Add sparse journal index to Raft log

kuujo:
This PR adds a sparse index to the Raft log to allow it to more quickly truncate/reset reader indexes. Slow seeking through large segments has caused significant blocking in Raft server threads, and using a sparse index allows us to much more efficiently navigate to specific portions of the Raft log without much overhead. The index is not thread safe because the Raft log is not thread safe. By default, the index stores .5% of entries.

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/proxy/impl/RaftProxyInvoker.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/proxy/impl/RaftProxyInvoker.java
@@ -128,7 +128,7 @@ final class RaftProxyInvoker {
         .withSession(state.getSessionId().id())
         .withSequence(state.getCommandRequest())
         .withOperation(operation)
-        .withIndex(Math.max(state.getResponseIndex(), state.getEventIndex()))
+        .withIndex(state.getResponseIndex())
         .build();
     invokeQuery(request, future);
   }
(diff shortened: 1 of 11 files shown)


protocols/raft/src/main/java/io/atomix/protocols/raft/proxy/impl/RaftProxyInvoker.java
```diff
@@ -128,7 +128,7 @@ private void invokeQuery(RaftOperation operation, CompletableFuture<byte[]> futu
         .withSession(state.getSessionId().id())
         .withSequence(state.getCommandRequest())
         .withOperation(operation)
-        .withIndex(Math.max(state.getResponseIndex(), state.getEventIndex()))
+        .withIndex(state.getResponseIndex())
```
kuujo: I'm not entirely sure how this got in here :-(

kuujo:
Ugh going to reopen this PR

protocols/raft/src/main/java/io/atomix/protocols/raft/proxy/impl/RaftProxyInvoker.java
johnou: https://github.com/atomix/atomix/pull/419

PR #424, opened 2018-02-11
Add sparse index to Raft log

kuujo:
This PR adds a sparse index to the Raft log to allow it to more quickly truncate/reset reader indexes. Slow seeking through large segments has caused significant blocking in Raft server threads, and using a sparse index allows us to much more efficiently navigate to specific portions of the Raft log without much overhead. The index is not thread safe because the Raft log is not thread safe. By default, the index stores .5% of entries.

### DIFF
--- a/storage/journal/src/test/java/io/atomix/storage/journal/index/SparseJournalIndexTest.java
+++ b/storage/journal/src/test/java/io/atomix/storage/journal/index/SparseJournalIndexTest.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2018-present Open Networking Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


kuujo:
Better

storage/journal/src/test/java/io/atomix/storage/journal/index/SparseJournalIndexTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+/**
+ * Sparse journal index test.
+ */
+public class SparseJournalIndexTest {
```
johnou: can we add a couple of tests for truncating too?

kuujo: 👍

johnou: Just noticed that the indexes are sequential, as this is a sparse index it might be worth skipping some to emulate snapshots?

kuujo: True

kuujo: Same tests on a new index with +100 added to all indexes

PR #436, opened 2018-03-02
Support configuring partition sizes in AtomixAgent

kuujo:


agent/src/main/java/io/atomix/agent/AtomixAgent.java
```diff
@@ -99,6 +99,18 @@ public File convert(ArgumentParser argumentParser, Argument argument, String val
...
+        .type(Integer.class)
+        .metavar("NUM")
+        .required(false)
+        .setDefault(7)
+        .help("The number of core partitions");
+    parser.addArgument("--data-partitions", "-dp")
+        .type(Integer.class)
+        .metavar("NUM")
+        .required(false)
+        .setDefault(71)
```
johnou: Is this a reasonable default?

kuujo: Just the same default as elsewhere. Probably needs to be reconsidered separately

PR #438, opened 2018-03-03
Add runtime shutdown hook to stop Atomix node

kuujo:


kuujo:
That would make sense if it were implemented in the `Atomix`, and maybe it should be, but this in `AtomixAgent` which is only run in an executable jar. There’s no need to have all that, aside from a lock inside the shutdown hook if it’s possible it’s called multiple times.

Are you saying it should be added to `Atomix` instead? I’m not sure whether that should be the responsibility of that class unless it’s optional.

johnou:
oh wow I totally missed that this was in the agent, fine as is.

kuujo:
I'm going to merge this but we might want to move it to `Atomix` some time

PR #439, opened 2018-03-05
Implement shutdown hook.

johnou:
@kuujo this is what I had in mind.


kuujo:
All for it, but a couple of changes:
* I'd make `doStop()` `synchronized`, then there's probably no need for a `shutdownLock` since `doStop()` is idempotent
* The builder methods need to follow the `with*` convention
* Replace the `AtomixAgent` changes with configuration of the `Atomix` instance

johnou:
@kuujo PTAL

PR #444, opened 2018-03-13
Enqueue entries to be applied to the Raft state machine

kuujo:
The current Raft implementation applies and recursively enqueues only one entry at a time. When the cluster is under high load, this can lead to entries being applied more slowly than they can be replicated, ultimately resulting in timeouts. This PR changes the implementation to enqueue all entries to be applied to the state machine as they're committed to ensure entries are applied concurrently with replication.


protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
```diff
@@ -306,95 +305,68 @@ public void applyAll(long index) {
...
   }
 
+  /**
+   * Applies all entries up to the given index.
+   *
+   * @param index the index up to which to apply entries
+   */
+  private void enqueueBatch(long index) {
+    while (lastEnqueued < index) {
+      enqueueIndex(++lastEnqueued);
```
johnou: @kuujo might be nicer to loop in the runnable to avoid context switching.

PR #445, opened 2018-03-15
Client retries

kuujo:
Cherry pick #443


protocols/raft/src/main/java/io/atomix/protocols/raft/proxy/impl/RaftProxyState.java
```diff
@@ -113,7 +114,18 @@ public long getSessionTimeout() {
...
       this.state = state;
+      if (state == PrimitiveProxy.State.SUSPENDED) {
+        if (suspendedTime == null) {
+          suspendedTime = System.currentTimeMillis();
+        }
+      } else {
+        suspendedTime = null;
+      }
       changeListeners.forEach(l -> l.accept(state));
+    } else if (this.state == PrimitiveProxy.State.SUSPENDED) {
```
johnou: Suspended state can be received twice?

kuujo: yep

PR #446, opened 2018-03-15
Enqueue entries to be applied to the state machine

kuujo:
Cherry pick #444

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
@@ -85,6 +85,7 @@ public class RaftServiceManager implements AutoCloseable {
   private final RaftLogReader reader;
   private final Map<Long, CompletableFuture> futures = Maps.newHashMap();
   private volatile CompletableFuture<Void> compactFuture;
+  private long lastEnqueued;
   private long lastCompacted;
 
   public RaftServiceManager(RaftContext raft, ThreadContext stateContext, ThreadContext compactionContext, ThreadContextFactory threadContextFactory) {
@@ -97,6 +98,7 @@ public class RaftServiceManager implements AutoCloseable {
     this.logger = ContextualLoggerFactory.getLogger(getClass(), LoggerContext.builder(RaftServer.class)
         .addValue(raft.getName())
         .build());
+    this.lastEnqueued = reader.getFirstIndex() - 1;
     scheduleSnapshots();
   }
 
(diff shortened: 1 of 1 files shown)


protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
```diff
@@ -305,94 +304,67 @@ public void applyAll(long index) {
...
+    enqueueBatch(index);
     return future;
   }
 
+  /**
+   * Applies all entries up to the given index.
+   *
+   * @param index the index up to which to apply entries
+   */
+  private void enqueueBatch(long index) {
```
johnou: @kuujo couldn't we optimize it like this?

```java
    private void enqueueBatch(long index) {
        raft.getThreadContext().execute(() -> {
            while (lastEnqueued < index) {
                applyIndex(++lastEnqueued);
            }
            return null;
        });
    }
```

protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
```diff
@@ -305,94 +304,67 @@ public void applyAll(long index) {
...
+    while (lastEnqueued < index) {
```
johnou: @kuujo couldn't we optimize it like this?

```java
    private void enqueueBatch(long index) {
        raft.getThreadContext().execute(() -> {
            while (lastEnqueued < index) {
                applyIndex(++lastEnqueued);
            }
            return null;
        });
    }
```

johnou: actually not exactly like that, not safe modifying lastEnqueued from the raft thread context..

kuujo: umm... yes that is a field not a local variable 🤷‍♂️

kuujo: No, this is specifically intended to enqueue each index as a separate item on the queue to ensure applying entries intersects with replication. The problem before _was_ effectively that many indexes were being read and applied in a single event, sometimes blocking other events that need to occur on the server thread.

johnou: How about controlled small batches instead of single entries?

kuujo: I actually did a couple of implementations, and 64KB batches were the first implementation. Don't recall why I went with this one now.

kuujo: Ahh, that's right. This one just performed best.

kuujo: Either way, I think we should keep this implementation for the fact that it's the one we're deploying to production (in Atomix 2.x) and thus have much more confidence in its correctness and efficiency. If any problems are encountered with queueing too many entries it can be revisited, but I'd like to keep parity between Atomix 2.1 and our production systems even though they're on 2.0 to ensure 2.1 is still benefiting from all the work we're putting the Raft implementation through.

PR #447, opened 2018-03-17
Implement timeouts in Raft test protocol client/server

kuujo:
This PR fixes the cause of frequent Raft test failures. The test failures are not due to any bug in the Raft implementation, but rather in an inconsistency between the implementation of the test protocol and the messaging service implementation. This PR adds timeouts to the Raft test protocol client and server to mimic the behavior of the messaging service that's used during normal operation.

### DIFF
--- a/protocols/raft/src/test/java/io/atomix/protocols/raft/protocol/TestRaftProtocol.java
+++ b/protocols/raft/src/test/java/io/atomix/protocols/raft/protocol/TestRaftProtocol.java
@@ -16,9 +16,14 @@
 package io.atomix.protocols.raft.protocol;
 
 import io.atomix.cluster.NodeId;
+import io.atomix.utils.concurrent.Scheduled;
+import io.atomix.utils.concurrent.ThreadContext;
 
+import java.time.Duration;
 import java.util.Collection;
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeoutException;
 
 /**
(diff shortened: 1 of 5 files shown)


protocols/raft/src/test/java/io/atomix/protocols/raft/protocol/TestRaftProtocol.java
```diff
@@ -16,20 +16,39 @@
...
+      Map<NodeId, TestRaftServerProtocol> servers,
+      Map<NodeId, TestRaftClientProtocol> clients,
+      ThreadContext context) {
     this.servers = servers;
     this.clients = clients;
+    this.context = context;
+  }
+
+  <T> CompletableFuture<T> scheduleTimeout(CompletableFuture<T> future) {
+    Scheduled scheduled = context.schedule(Duration.ofMillis(100), () -> {
```
johnou: No response under 100 ms results in timeout?

kuujo: They're all local in tests anyways, so I suspect a 100ms response time from another thread would be indicative of a problem anyways.

johnou: Or an overloaded Travis node ;)

kuujo: touché

PR #449, opened 2018-03-20
Ensure tryLock attempts are rejected during network partitions

kuujo:
This PR is a significant improvement to the behavior of `DistributedLock` primitives. These changes come out of work on tolerating network partitions in ONOS. They primarily revolve around ensuring locks attempt timeouts can be controlled by clients and that state is cleaned up when clients' lock attempts time out. Significant documentation of the changes is provided inline.

### DIFF
--- a/core/src/main/java/io/atomix/core/Atomix.java
+++ b/core/src/main/java/io/atomix/core/Atomix.java
@@ -80,6 +80,7 @@ import io.atomix.utils.Managed;
 import io.atomix.utils.concurrent.Futures;
 import io.atomix.utils.concurrent.SingleThreadContext;
 import io.atomix.utils.concurrent.ThreadContext;
+import io.atomix.utils.concurrent.Threads;
 import io.atomix.utils.net.Address;
 import io.atomix.utils.net.MalformedAddressException;
 import org.slf4j.Logger;
@@ -91,6 +92,8 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/atomix/core/Atomix.java
```diff
@@ -263,6 +269,7 @@ public boolean isRunning() {
         .thenComposeAsync(v -> metadataService.stop(), context)
         .thenComposeAsync(v -> messagingService.stop(), context)
         .thenRunAsync(() -> {
+          executorService.shutdown();
```
johnou: graceful executor shutdown?

```java
executorService.shutdown(); // Disable new tasks from being submitted
try {
    // Wait a while for existing tasks to terminate
    if (!executorService.awaitTermination(10, TimeUnit.SECONDS)) {
        logger.info("Timeout elapsed before termination, forcing shutdown");
        List<Runnable> tasksAwaitingExecution = executorService.shutdownNow(); // Cancel currently executing tasks
        logger.info("Tasks awaiting execution after forced shutdown: " + tasksAwaitingExecution.size());
        // Wait a while for tasks to respond to being cancelled
        if (!executorService.awaitTermination(10, TimeUnit.SECONDS)) {
            logger.error("Pool did not terminate");
        }
    }
} catch (InterruptedException ie) {
    logger.error("Exception occurred while shutting down thread pool", ie);
    // (Re-)Cancel if current thread also interrupted
    executorService.shutdownNow();
    // Preserve interrupt status
    Thread.currentThread().interrupt();
}
logger.info("Thread pool shutdown complete");
```

johnou: Only potential problem I see is when locks are created with a high timeout the JVM might hang on shutdown while waiting for those? Alternatively the executor could be configured to create daemon threads?

johnou: @kuujo since we already stopped messaging before this couldn't we just invoke shutdownNow?

johnou: @kuujo poke

kuujo: I don't have the motivation to fix the merge conflicts ATM :-P

kuujo: okay did

johnou:
Lgtm

Issue #430, opened 2018-03-01
long leadership election 


lukasz-plewnia:
Hello guys

First of all, nice work so far!

Code: [https://gist.github.com/lukasz-plewnia/96cc21494b3821b35ab0b1d43cda36fa](https://gist.github.com/lukasz-plewnia/96cc21494b3821b35ab0b1d43cda36fa)

I'm recently testing atomix (ver. 2.1.0-SNAPSHOT) and I ran into this issue.
I'm running local atomix cluster with local client. When I kill the current leader node, the client gets notification about the leader change in either ~1 second, ~7 seconds or it does not get the notification at all.

Am I doing it wrong or it happens to be that slow?
Thanks

johnou:
@lukasz-plewnia snapshot as in head of master? Would you please check with 2.1.0-beta2?

lukasz-plewnia:
Exactly, local build as in head of master.
Same thing happens with version 2.1.0-beta2.

johnou:
@kuujo ptal

kuujo:
Sorry I’ve been out of town for several weeks. Awesome reproducer! Looks correct to me. I’ll run it and see what I can come up with.

BTW the differences in leader election times are likely the result of the state of the cluster when a node is stopped. If the node is a leader for the Raft partition that store’s the election, a Raft leader change sometimes has to occur before the actual `LeaderElector` failure can be detected and handled. And because Raft leader elections can take an arbitrary amount of time, `LeaderElector` session times also have to be reset after the election. Also, sometimes failures may be detected via phi accrual failure detectors (which are usually faster) and at other times they may be detected via Raft keep-alives. This again just depends on the state of the cluster when the failure occurs.

kuujo:
I created a modified version of the same gist just for reference sake:
https://gist.github.com/kuujo/82a565e0e73b331d9c17870bee820a58

The only changes here are really that it blocks waiting for a leader change event to be received after the leader is killed.

The cause of the strange election times is indeed just a product of failure detectors. Often, in this test the cluster has not been running long enough for failure detectors to be populated for fast detection. So, the Raft session (~7 seconds) has to expire before the leader change can occur. At other times, the failure detectors are populated and the failure is detected in a second.

The bug that is preventing the leader change in some cases seems to be an exception thrown in the `DefaultClusterMessagingService` that's not handled by Raft clients. When `server1` becomes unavailable, for some reason it's removed from the `ClusterService` which is likely a bug.

[...]

PR #451, opened 2018-03-22
Return ConnectException when peer is unavailable

kuujo:
The `DefaultCLusterMessagingService` looks up nodes via the `ClusterService` and throws an `IllegalArgumentException` if a message is sent to an unknown node. Instead, it should return `ConnectException` to ensure Raft clients - which are designed to handle connection issues - properly handle the error and look for another node to connect to. This is the cause of #430.

### DIFF
--- a/cluster/src/main/java/io/atomix/cluster/messaging/impl/DefaultClusterMessagingService.java
+++ b/cluster/src/main/java/io/atomix/cluster/messaging/impl/DefaultClusterMessagingService.java
@@ -27,6 +27,7 @@ import io.atomix.utils.concurrent.Futures;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.net.ConnectException;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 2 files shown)


cluster/src/main/java/io/atomix/cluster/messaging/impl/DefaultClusterMessagingService.java
```diff
@@ -133,13 +134,17 @@ public DefaultClusterMessagingService(ClusterService cluster, MessagingService m
...
+      return Futures.exceptionalFuture(new ConnectException());
+    }
     return messagingService.sendAsync(node.endpoint(), subject, payload);
   }
 
   private CompletableFuture<byte[]> sendAndReceive(String subject, byte[] payload, NodeId toNodeId) {
     Node node = cluster.getNode(toNodeId);
-    checkArgument(node != null, "Unknown nodeId: %s", toNodeId);
+    if (node == null) {
+      return Futures.exceptionalFuture(new ConnectException());
```
johnou: Could omit the stacktrace for performance?

kuujo: yep good idea

kuujo: fixed

kuujo:
Oops there's a rogue file in this PR

johnou:
Lgtm

PR #453, opened 2018-03-23
Support revisions in Raft state machines

kuujo:
This PR implements support for revisions in Raft state machines.

Revisions are essentially versioning for Raft state machines with additional features for optionally propagating some state changes. We use revisions when upgrading ONOS controllers. During rolling upgrades, new primitive revisions can be initialized from the existing primitive state, and changes may or may not be propagated from older revisions to newer revisions based on the requirements of specific applications.

The change supports three `PropagationStrategy`s for revisions:
* `NONE` completely isolates changes to each version
* `VERSION` sets all prior revisions to read-only and supports writes only on the latest revision
* `PROPAGATE` propagates changes on older revisions to newer revisions

When changes are propagated from an older state machine revision to a newer revision, all basic features - including events - continue to operate as normal. The operation behaves just as a normal command on the state machine would.

Revisions are implemented simply by taking temporary snapshots of an existing state machine, creating a new instance of the same state machine, and installing the snapshot on the new state machine.

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
@@ -20,14 +20,16 @@ import com.google.common.primitives.Longs;
 import io.atomix.protocols.raft.RaftException;
 import io.atomix.protocols.raft.RaftServer;
 import io.atomix.protocols.raft.cluster.MemberId;
+import io.atomix.protocols.raft.service.PropagationStrategy;
 import io.atomix.protocols.raft.service.RaftService;
 import io.atomix.protocols.raft.service.ServiceId;
+import io.atomix.protocols.raft.service.ServiceRevision;
 import io.atomix.protocols.raft.service.ServiceType;
 import io.atomix.protocols.raft.service.impl.DefaultServiceContext;
+import io.atomix.protocols.raft.session.RaftSession;
 import io.atomix.protocols.raft.session.RaftSessionMetadata;
 import io.atomix.protocols.raft.session.SessionId;
 import io.atomix.protocols.raft.session.impl.RaftSessionContext;
-import io.atomix.protocols.raft.storage.log.RaftLog;
 import io.atomix.protocols.raft.storage.log.RaftLogReader;
 import io.atomix.protocols.raft.storage.log.entry.CloseSessionEntry;
 import io.atomix.protocols.raft.storage.log.entry.CommandEntry;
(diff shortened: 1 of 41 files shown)


protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
```diff
@@ -593,51 +603,101 @@ private boolean completeSnapshot(long index) {
...
+      ServiceType serviceType,
+      String serviceName,
+      int revision,
+      PropagationStrategy propagationStrategy) {
+    if (revision == 0) {
+      DefaultServiceContext service = raft.getServices().getCurrentRevision(serviceName);
+      if (service != null) {
+        return service;
+      } else {
+        revision = 1;
```
johnou: why 1? will it correct itself after talking with the other services?

protocols/raft/src/main/java/io/atomix/protocols/raft/service/impl/DefaultServiceContext.java
```diff
@@ -447,6 +485,27 @@ private OperationResult sequenceCommand(long index, long sequence, RaftSessionCo
...
+    if (nextRevision != null) {
+      switch (nextRevision.revision().propagationStrategy()) {
+        case VERSION:
+          return OperationResult.failed(
+              index,
+              session.getEventIndex(),
+              new RaftException.ReadOnly(serviceName() + " revision " + revision.revision() + " is read-only"));
+        case PROPAGATE:
+          locked = true;
+          nextRevision.applyCommand(index, sequence, timestamp, operation, session);
```
johnou: try finally and unlock incase applycommand throws exception?

protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/SnapshotDescriptor.java
```diff
@@ -167,6 +192,28 @@ public Builder withTimestamp(long timestamp) {
...
+    }
+
+    /**
+     * Sets the snapshot propagation strategy.
+     *
+     * @param synchronization the snapshot propagation strategy
+     * @return the snapshot builder
+     */
+    public Builder withPropagation(PropagationStrategy synchronization) {
+      buffer.writeByte(21, synchronization.ordinal());
```
johnou: magic numbers?

protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
kuujo: The default revision if no revision number is specified and no revision exists is 1. If a revision exists then the latest is used.

kuujo: This basically creates a revision of `1` if the service doesn't exist or returns the latest revision if one does exist if `0` (the default) is provided

kuujo:
Just gotta fix some test failures

kuujo:
This is going to be complicated to get into 2.1. I think it's going to look a lot different there. I'll try to get around to it soon though.

PR #458, opened 2018-03-25
Pooled IO objects.

johnou:
This pull request introduces pooled IO objects which sacrifices memory for speed (increase of 120000 op/s).

```
Atomix (before)

Benchmark                                         Mode  Cnt       Score      Error  Units
SingleNodeBenchmark.requestThroughput            thrpt  100  217284.327 ± 1274.281  ops/s
SingleNodeBenchmark.avgRequestTime_batched        avgt  100       5.976 ±    0.029  us/op
SingleNodeBenchmark.avgRequestTime_singleThread   avgt  100      19.552 ±    0.120  us/op

Atomix (after)

Benchmark                                         Mode  Cnt       Score      Error  Units
SingleNodeBenchmark.requestThroughput            thrpt  100  345203.644 ± 3027.675  ops/s
SingleNodeBenchmark.avgRequestTime_batched        avgt  100       4.729 ±    0.030  us/op
SingleNodeBenchmark.avgRequestTime_singleThread   avgt  100      17.532 ±    0.374  us/op

JGroups

Benchmark                                         Mode  Cnt       Score      Error  Units
SingleNodeBenchmark.requestThroughput            thrpt  100  342470.417 ± 2624.415  ops/s
SingleNodeBenchmark.avgRequestTime_batched        avgt  100       3.340 ±    0.029  us/op
SingleNodeBenchmark.avgRequestTime_singleThread   avgt  100      22.910 ±    0.149  us/op
```

With these changes Atomix is on-par with JGroups throughput (within error margin).

### DIFF
--- a/utils/src/main/java/io/atomix/utils/serializer/BufferAwareByteArrayOutputStream.java
+++ b/utils/src/main/java/io/atomix/utils/serializer/BufferAwareByteArrayOutputStream.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2014-present Open Networking Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.atomix.utils.serializer;
+
+import java.io.ByteArrayOutputStream;
+
+/**
+ * Exposes protected byte array length in {@link ByteArrayOutputStream}.
+ */
+final class BufferAwareByteArrayOutputStream extends ByteArrayOutputStream {
+
+    BufferAwareByteArrayOutputStream(int size) {
+        super(size);
+    }
+
+    int getBufferSize() {
+        return buf.length;
+    }
+}
--- a/utils/src/main/java/io/atomix/utils/serializer/KryoIOPool.java
+++ b/utils/src/main/java/io/atomix/utils/serializer/KryoIOPool.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2014-present Open Networking Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.atomix.utils.serializer;
+
+import java.lang.ref.SoftReference;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.function.Function;
+
+abstract class KryoIOPool<T> {
+
+    private final Queue<SoftReference<T>> queue = new ConcurrentLinkedQueue<>();
+
+    private T borrow(final int bufferSize) {
+        T element;
+        SoftReference<T> reference;
+        while ((reference = queue.poll()) != null) {
+            if ((element = reference.get()) != null) {
+                return element;
+            }
+        }
+        return create(bufferSize);
+    }
+
+    protected abstract T create(final int bufferSize);
+
+    protected abstract boolean recycle(final T element);
+
+    <R> R run(final Function<T, R> function, final int bufferSize) {
+        final T element = borrow(bufferSize);
+        try {
+            return function.apply(element);
+        } finally {
+            if (recycle(element)) {
+                queue.offer(new SoftReference<>(element));
+            }
+        }
+    }
+}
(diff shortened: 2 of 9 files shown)


cluster/src/main/java/io/atomix/cluster/messaging/impl/ClusterMessage.java
```diff
@@ -1,171 +0,0 @@
...
-
-import java.nio.ByteBuffer;
-import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
-import java.util.Objects;
-
-/**
- * Base message for cluster-wide communications.
- */
-public class ClusterMessage {
```
johnou: Seemed pointless to keep this, only thing that seemed to be in use was the payload.

utils/src/main/java/io/atomix/utils/serializer/KryoIOPool.java
```diff
@@ -0,0 +1,35 @@
+package io.atomix.utils.serializer;
+
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.function.Function;
+
+class KryoIOPool<T> {
```
johnou: only "risks" I see here is that the pool queue is unbounded and we may retain elements with large bufferSizes.

utils/src/main/java/io/atomix/utils/serializer/KryoNamespace.java
```diff
@@ -296,13 +310,17 @@ public KryoNamespace populate(int instances) {
...
-      kryo.writeClassAndObject(out, obj);
-      out.flush();
-      return outputStream.toByteArray();
-    });
+    return outputPool.run(output -> {
+      try {
+        return kryoPool.run(kryo -> {
+          kryo.writeClassAndObject(output, obj);
+          output.flush();
+          return output.getStream().toByteArray();
```
johnou: no risk of data corruption here as toByteArray returns a new byte array.

utils/src/main/java/io/atomix/utils/serializer/KryoNamespace.java
```diff
@@ -78,10 +78,24 @@
...
+    @Override
+    protected Input create(final int bufferSize) {
+      return new Input(4096);
+    }
+  };
+
+  private final KryoIOPool<ByteArrayOutput> outputPool = new KryoIOPool<ByteArrayOutput>() {
+    @Override
+    protected ByteArrayOutput create(final int bufferSize) {
+      return new ByteArrayOutput(4096, 4096, new ByteArrayOutputStream(bufferSize));
```
johnou: magic numbers from kryo Output / Input constructors.

utils/src/main/java/io/atomix/utils/serializer/KryoIOPool.java
johnou: @NathanSweet have you considered pooling Input and Output objects in Kryo?

johnou: bytearrayoutputstreams with buffer exceeding MAX_BUFFER_SIZE are no longer retained in the pool.

utils/src/main/java/io/atomix/utils/serializer/BufferAwareByteArrayOutputStream.java
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Exposes protected byte array length in {@link ByteArrayOutputStream}.
+ */
+final class BufferAwareByteArrayOutputStream extends ByteArrayOutputStream {
+
+    BufferAwareByteArrayOutputStream(int size) {
+        super(size);
+    }
+
+    int getBufferSize() {
```
johnou: buf never shrinks, only grows, and flush resets the size so we need to check the size like this.

utils/src/main/java/io/atomix/utils/serializer/KryoIOPool.java
NathanSweet: I think it's application level logic to manage Input and Output allocations.

johnou: To me seemed more logical to keep it with KyroPool but thanks for the feedback.

NathanSweet: That is true. FWIW, it wasn't me who added KryoPool. ;)

cluster/src/main/java/io/atomix/cluster/messaging/impl/ClusterMessage.java
kuujo: Yeah I think this class was actually left over from an old version of the ONOS API

kuujo:
Looks good to me. I think we just need to throw the standard license/copyright headers in all the files and some basic documentation.

I actually need to setup checkstyle to enforce these things. I added it to the pom but never actually added any rules or tried to get it to work in Travis.

kuujo:
Also would like to see some tests on the pool. The low level I/O stuff is pretty well covered by tests and we should try to keep it that way.

utils/src/main/java/io/atomix/utils/serializer/KryoIOPool.java
johnou: @kuujo maybe we could wrap the pool objects with a SoftReference like we already use for kryopool to keep it consistent?

utils/src/main/java/io/atomix/utils/serializer/KryoOutputPool.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.atomix.utils.serializer;
+
+import com.esotericsoftware.kryo.io.Output;
+
+class KryoOutputPool extends KryoIOPool<Output> {
+
+    private static final int MAX_BUFFER_SIZE = 768 * 1024;
```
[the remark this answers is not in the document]
johnou: Addressed.

kuujo:
Just `PrimaryBackupDocumentTreeTest` failing again. Not sure what's up will that

Issue #455, opened 2018-03-25
Support partition member groups


kuujo:
Add support for host/rack/zone/group aware member groups for partitions. This would allow primary-backup partitions to be replicated across physical locations. Each `Node` can be assigned a:
• `host`
• `rack`
• `zone`
• `group`

Additionally, custom member groups can be defined for a partition group.

PR #463, opened 2018-03-28
Partition member groups

kuujo:
#455 

This PR implements support for partition member groups. `MemberGroup`s are used to ensure replicas are distributed across multiple physical systems. Groups can be created based on new `Node` fields - `zone()`/`rack()`/`host()`. It adds a custom primitive state machine that handles primary election for the primary-backup protocol to ensure partitions are evenly distributed across the cluster. The member groups can be configured via a `MemberGroupStrategy`:

```java
PartitionGroup partitionGroup = PrimaryBackupPartitionGroup.builder("foo")
  .withNumPartitions(32)
  .withMemberGroupStrategy(MemberGroupStrategy.RACK_AWARE)
  .build();
```


kuujo:
There’s a rogue `atomix.log` file in this PR. Need to just add `.log` to `.gitignore`

kuujo:
Also, this PR is on top of the `CORE` PR so I suppose we should get that merged and then rebase this.

PR #464, opened 2018-03-29
Add support for DATA node only clusters

kuujo:
This PR is a WORK IN PROGRESS that adds support for forming clusters only from `DATA` and `CLIENT` nodes. This can be done using a couple of methods:
* Provide one or more node `Endpoint`s to join
* Enable multicast discovery

A `BootstrapMetadataService` is added to provide information about nodes to attempt to join, and the `ClusterMetadataService` was renamed to `CoreMetadataService` and is now responsible for providing a static configuration for `CORE` nodes. A `BroadcastService` was added as a peer to `MessagingService` to handle the multicast protocol, and the `BroadcastService` is used within `ClusterService` to learn about other nodes via multicast. Once a node is discovered via multicast, the normal TCP-based failure detection proceeds.

Multicast is enabled via the `Atomix` builder and can be provided a multicast group address:

```java
Atomix atomix = Atomix.builder()
  .withLocalNode(Node.builder("foo")
    .withType(Node.Type.DATA)
    .withEndpoint(Endpoint.from("127.0.0.1", 1234))
    .build())
  .withMulticastEnabled()
  .withMulticastEndpoint(Endpoint.from("230.0.0.1", 5678))
  .build();
```

I'm a little curious if we might want to rename `Endpoint` now or at least create a base `Address` class that can be used in the multicast address.

The current implementation is pretty ugly. I suspect a lot of cleanup and testing will need to be done. The implementation of a non-Raft based `SessionId` generator is pretty lazy, and better distributed systems algorithms for implementing globally unique counters exist. The `PrimaryElection` term could benefit from the same algorithm.


kuujo:
Should be good to go for now. More work will be done in additional PRs.

PR #465, opened 2018-03-30
Pending command commit after leader change

kuujo:
This PR fixes a bug that can create a livelock in a Raft client's session immediately following a leader change.

The bug requires a series of events involving queueing of commands inside Raft servers and retrying commands on clients:
* Client sends command `1` to the leader for term `1`
* The leader for term `1` replicates the command to a majority of the cluster and then crashes
* A new leader is elected for term `2`
* The new leader commits entries from term `1`
* The client retries command `1` on the leader for term `2`
* The leader for term `2` queues the command as a `PendingCommand` since its sequence number does not align (hasn't yet applied entries from the prior term)
* The leader for term `2` then applies command `1`, thus incrementing the client's `requestSequence` beyond that of command `1`
* The client again retries command `1`, and the leader for term `2` again responds with the `PendingCommand` future

The last point is where the livelock is. Because the `requestSequence` number has been incremented beyond that of command `1`, the `PendingCommand` is never committed, nor are any later pending commands for the session. This PR ensures that commands up to `requestSequence` are drained.


kuujo:
Opened on the wrong branch damnit

PR #467, opened 2018-03-31
Replace primitive configurations with PrimitiveProtocol requirement

kuujo:
This PR removes confusing `Replication`/`Consistency`/etc configurations from primitive builders and replaces them only with protocol configurations.


primitive/src/main/java/io/atomix/primitive/partition/impl/DefaultPartitionService.java
```diff
@@ -49,13 +49,14 @@ public PartitionGroup getPartitionGroup(String name) {
 
   @Override
   @SuppressWarnings("unchecked")
-  public Collection<PartitionGroup> getPartitionGroups() {
+  public Collection<PartitionGroup<?>> getPartitionGroups() {
```
johnou: Why wildcard?

kuujo: Without the wildcard I have to do a lot of casting when using the streams API to iterate through groups and partitions

PR #469, opened 2018-04-01
Rename Endpoint class to Address

kuujo:
Renames the `Endpoint` class to `Address` as it will now be used to represent more than just endpoints.

### DIFF
--- a/agent/src/main/java/io/atomix/agent/AtomixAgent.java
+++ b/agent/src/main/java/io/atomix/agent/AtomixAgent.java
@@ -18,10 +18,10 @@ package io.atomix.agent;
 import io.atomix.cluster.Node;
 import io.atomix.cluster.NodeId;
 import io.atomix.core.Atomix;
-import io.atomix.utils.net.Address;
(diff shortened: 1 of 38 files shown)


jplock:
Could this be replaced with guava’s HostAndPort?

johnou:
@jplock no, guava will be ripped out at some point.

agent/src/main/java/io/atomix/agent/AtomixAgent.java
```diff
@@ -165,7 +165,7 @@ public static void main(String[] args) throws Exception {
     }
   }
 
-  static String[] parseAddress(String address) {
+  static String[] parseInfo(String address) {
     String[] parsed = address.split(":");
     if (parsed.length > 3) {
       throw new IllegalArgumentException("Malformed address " + address);
```
johnou: Trim strings after split?

kuujo: If there's whitespace in the string then it won't be parsed by the command line parser as a single argument and will thus be malformed anyways

PR #471, opened 2018-04-02
Improve efficiency of resolving Raft log conflicts

kuujo:
This PR improves the efficiency of resolving differences between a Raft leader's and follower's logs by providing additional information in rejected `AppendResponse`s. This is done by tracking the indexes at which each term starts in the Raft log and returning the first index of the rejected term to the leader. This information is then used by the leader to skip conflicting entries in that term, thus significantly reducing the amount of time it takes to find the consistent portion of the follower's logs.

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/protocol/AppendResponse.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/protocol/AppendResponse.java
@@ -39,12 +39,14 @@ public class AppendResponse extends AbstractRaftResponse {
   private final long term;
   private final boolean succeeded;
   private final long lastLogIndex;
+  private final long lastLogTerm;
 
-  public AppendResponse(Status status, RaftError error, long term, boolean succeeded, long lastLogIndex) {
+  public AppendResponse(Status status, RaftError error, long term, boolean succeeded, long lastLogIndex, long lastLogTerm) {
     super(status, error);
     this.term = term;
     this.succeeded = succeeded;
     this.lastLogIndex = lastLogIndex;
+    this.lastLogTerm = lastLogTerm;
   }
 
   /**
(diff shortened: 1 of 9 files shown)


protocols/raft/src/main/java/io/atomix/protocols/raft/roles/AbstractAppender.java
```diff
@@ -293,24 +292,26 @@ protected void updateMatchIndex(RaftMemberContext member, AppendResponse respons
...
-   */
-  protected void resetMatchIndex(RaftMemberContext member, AppendResponse response) {
-    if (response.lastLogIndex() < member.getMatchIndex()) {
-      member.setMatchIndex(response.lastLogIndex());
-      log.trace("Reset match index for {} to {}", member, member.getMatchIndex());
-    }
-  }
-
   /**
    * Resets the next index when a response fails.
```
johnou: Update javadoc?

protocols/raft/src/main/java/io/atomix/protocols/raft/protocol/AppendResponse.java
```diff
@@ -86,7 +97,8 @@ public boolean equals(Object object) {
       return response.status == status
           && response.term == term
           && response.succeeded == succeeded
-          && response.lastLogIndex == lastLogIndex;
+          && response.lastLogIndex == lastLogIndex
```
johnou: update equals too?

johnou: @kuujo poke, no conflicts here ;p

johnou:
@kuujo is this something we need to address in 3.0?

kuujo:
Yeah, it just hasn't been well tested because it's been low priority. I'll
eventually get to it and clean up and test it.

kuujo:
We can probably just close this for the time

PR #473, opened 2018-04-04
Add support for file-based configurations

kuujo:
This PR adds support for YAML and JSON configuration files.

It's implemented by adding Java beans for all types that currently have builders. Configuration files are parsed by a `ConfigProvider`. The built-in `ConfigProvider` is based on Jackson, but additional providers can be included by registering the class via the service loader pattern.

In addition to the `ConfigProvider` being pluggable, numerous other APIs like `PrimitiveProtocol` and `PartitionGroup` now support the service loader pattern as well. This is used by the Jackson `ConfigProvider` to dynamically register polymorphic types with the serializer.

The objects that can be configured via configuration files are:
* `Atomix`
* The cluster
* The core/bootstrap nodes
* The primitive types
* Specific named primitives
* Partition groups
* Partition member groups
* Primitive protocols
* Serializer registrations

This allows an `Atomix` instance and all its primitives to be entirely configured via a configuration file. To configure an Atomix instance, simply construct it with a `String` file name or `File` object:

```java
Atomix atomix = new Atomix("atomix.json");
```

Additionally, configuration files and builders can be mixed:

```java
Atomix atomix = Atomix.builder("atomix.json")
  .withLocalNode(Node.builder("foo")
    .withAddress("localhost:1234")
    .build())
  .build();
```

The Atomix agent has also been updated to support configuration files:
```
java -jar atomix-agent.jar -c atomix.yaml
```

As well as a mixture of arguments:
```
java -jar atomix-agent.jar foo:localhost:1234 -c atomix.yaml
```

This is particularly useful for sharing a configuration file among all the nodes in a cluster and passing only local arguments on the command line or to the `Atomix` builder.

Since primitives can now be configured via configuration files, I also added non-builder methods for getting singleton primitive instances:

```java
ConsistentMap<String, String> map = atomix.getConsistentMap("foo");
```

The map can be configured via a configuration file:

```yaml
cluster:
  name: test
  core:
    nodes:
      - id: node1
        type: CORE
        address: localhost:5000
      - id: node2
        type: CORE
        address: localhost:5001
      - id: node3
        type: CORE
        address: localhost:5002
partition-groups:
  - type: raft
    name: core
    storage-level: MEMORY
    partitions: 7
    partition-size: 3
    data-directory: target/test-logs/core
primitives:
  foo:
    type: CONSISTENT_MAP 
    null-values: false
    protocol:
      type: raft
      read-consistency: LINEARIZABLE
      max-retries: 5
```


kuujo:
FYI I think we should modify the deserializer configuration some more to make enum values case-insensitive

kuujo:
Also need to support custom primitive type configurations

kuujo:
This is obviously a work in progress, but I'm going to merge it into `master` so the remaining work can be done in smaller reviewable PRs.

PR #474, opened 2018-04-04
Replace "==" with "equals"

tanbangcheng:


tanbangcheng:
it has been fixed

kuujo:
Good catch!

PR #478, opened 2018-04-06
Support interpolating environment variables and system properties in config

kuujo:
This PR adds support for interpolating environment variables and system properties in YAML configuration files, e.g.:

```yaml
partition-groups:
  - name: test
     type: raft
     partitions: ${NUM_RAFT_PARTITIONS}
     partition-size: ${raft.partition.size}
```

### DIFF
--- a/config/src/main/java/io/atomix/core/config/jackson/JacksonConfigProvider.java
+++ b/config/src/main/java/io/atomix/core/config/jackson/JacksonConfigProvider.java
@@ -16,6 +16,9 @@
 package io.atomix.core.config.jackson;
 
 import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.ObjectCodec;
+import com.fasterxml.jackson.core.io.IOContext;
+import com.fasterxml.jackson.core.util.BufferRecycler;
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.MapperFeature;
@@ -25,6 +28,7 @@ import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import com.fasterxml.jackson.dataformat.yaml.YAMLParser;
 import io.atomix.core.config.ConfigProvider;
 import io.atomix.primitive.PrimitiveConfig;
 import io.atomix.primitive.PrimitiveProtocolConfig;
@@ -38,9 +42,13 @@ import io.atomix.utils.ConfigurationException;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.Reader;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * Jackson configuration provider.
@@ -75,7 +83,7 @@ public class JacksonConfigProvider implements ConfigProvider {
   }
 
   private <C extends Config> C loadYaml(File file, Class<C> type) {
-    ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+    ObjectMapper mapper = new ObjectMapper(new InterpolatingYamlFactory());
     setupObjectMapper(mapper);
     try {
       return mapper.readValue(file, type);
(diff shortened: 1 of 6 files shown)


johnou:
@kuujo maybe we should take the same approach as log4j2?

https://logging.apache.org/log4j/2.x/manual/configuration.html

Prefix	Context

env	System environment variables. The formats are ${env:ENV_NAME} and ${env:ENV_NAME:-default_value}.

sys	System properties. The formats are ${sys:some.property} and ${sys:some.property:-default_value}.

johnou:
I think I remember reading somewhere it was for security reasons.

config/src/main/java/io/atomix/core/config/jackson/JacksonConfigProvider.java
```diff
@@ -179,4 +188,51 @@ public PrimitiveConfigDeserializer() {
...
+    }
+
+    @Override
+    public String getValueAsString(String defaultValue) throws IOException {
+      String value = super.getValueAsString(defaultValue);
+      return value != null ? interpolateString(value) : null;
+    }
+
+    private String interpolateString(String value) {
+      Pattern pattern = Pattern.compile(PATTERN);
```
johnou: why not compile once in field?

config/src/main/java/io/atomix/core/config/jackson/JacksonConfigProvider.java
```diff
@@ -179,4 +188,51 @@ public PrimitiveConfigDeserializer() {
...
+      Matcher matcher = pattern.matcher(value);
+      while (matcher.find()) {
+        String name = matcher.group(1);
+        String replace = System.getenv(name);
```
johnou: I would recommend splitting these as suggested in my other commit.

kuujo:
That seems more clear, sure

kuujo:
I tried to consider what the security concerns might be here, but this parser is currently only used to parse local configuration files. I’d imagine this can become a security risk if we use the same parser in the REST API (I plan to use it but remove interpolation in that case) or support pulling from an external URL (entirely possible future feature). We’d probably want to disable it for the latter case as well and ensure they can only be used for local configuration files. But the usage of the configuration parser can especially become vulnerable I think with respect to injecting classes into the serializer configuration, which should probably also be disabled for the REST API since there’s no reason for a user of that API to register classes anyways.

config/src/main/java/io/atomix/core/config/jackson/JacksonConfigProvider.java
```diff
@@ -179,4 +188,54 @@ public PrimitiveConfigDeserializer() {
...
+
+  private static class InterpolatingYamlFactory extends YAMLFactory {
+    @Override
+    protected YAMLParser _createParser(InputStream in, IOContext ctxt) throws IOException {
+      return new InterpolatingYamlParser(ctxt, _getBufferRecycler(), _parserFeatures, _yamlParserFeatures, _objectCodec, _createReader(in, null, ctxt));
+    }
+  }
+
+  private static class InterpolatingYamlParser extends YAMLParser {
+    private static final String SYS_PATTERN = "\\$\\{sys:([A-Za-z0-9-_.]+)\\}";
```
johnou: I meant something like this

```java
private final Pattern sysPattern = Pattern.compile("\\$\\{sys:([A-Za-z0-9-_.]+)\\}");
private final Pattern envPattern = Pattern.compile("\\$\\{env:([A-Za-z0-9_]+)\\}");
```

config/src/test/java/io/atomix/core/config/jackson/JacksonConfigProviderTest.java
```diff
@@ -47,4 +48,24 @@ public void testYaml() throws Exception {
     assertEquals("test", config.getClusterConfig().getName());
     assertEquals(1, config.getPrimitives().get("foo").getSerializerConfig().getTypes().size());
   }
+
+  @Test
+  @Ignore
```
johnou: why ignore? add env.yaml to src/test/resources?

kuujo:
The regex is not compiled because the string is reused inside the `interpolate` method

config/src/test/java/io/atomix/core/config/jackson/JacksonConfigProviderTest.java
kuujo: Can't set an environment variable in a junit test

johnou:
so what if it's reused?

```java
Index: config/src/main/java/io/atomix/core/config/jackson/JacksonConfigProvider.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- config/src/main/java/io/atomix/core/config/jackson/JacksonConfigProvider.java	(date 1522985409000)
+++ config/src/main/java/io/atomix/core/config/jackson/JacksonConfigProvider.java	(revision )
@@ -197,8 +197,8 @@
   }
 
   private static class InterpolatingYamlParser extends YAMLParser {
-    private static final String SYS_PATTERN = "\\$\\{sys:([A-Za-z0-9-_.]+)\\}";
-    private static final String ENV_PATTERN = "\\$\\{env:([A-Za-z0-9_]+)\\}";
+    private final Pattern sysPattern = Pattern.compile("\\$\\{sys:([A-Za-z0-9-_.]+)\\}");
+    private final Pattern envPattern = Pattern.compile("\\$\\{env:([A-Za-z0-9_]+)\\}");
 
     InterpolatingYamlParser(IOContext ctxt, BufferRecycler br, int parserFeatures, int formatFeatures, ObjectCodec codec, Reader reader) {
       super(ctxt, br, parserFeatures, formatFeatures, codec, reader);
@@ -222,13 +222,13 @@
     }
 
     private String interpolateString(String value) {
-      value = interpolate(value, SYS_PATTERN, name -> System.getProperty(name, ""));
-      value = interpolate(value, ENV_PATTERN, name -> System.getenv(name));
+      value = interpolate(value, sysPattern, name -> System.getProperty(name, ""));
+      value = interpolate(value, envPattern, name -> System.getenv(name));
       return value;
     }
 
-    private String interpolate(String value, String pattern, Function<String, String> supplier) {
-      Matcher matcher = Pattern.compile(pattern).matcher(value);
+    private String interpolate(String value, Pattern pattern, Function<String, String> supplier) {
+      Matcher matcher = pattern.matcher(value);
       while (matcher.find()) {
         String name = matcher.group(1);
         String replace = supplier.apply(name);
```

kuujo:
I knew what you meant... but it's actually not reused anyways. I was thinking about the `Pattern` that has to be dynamically compiled inside that method.

johnou:
oh right io.atomix.core.config.jackson.JacksonConfigProvider.InterpolatingYamlParser#getValueAsString(java.lang.String) is only called once in the entire lifecycle of the parser object?

config/src/main/java/io/atomix/core/config/jackson/JacksonConfigProvider.java
```diff
@@ -179,4 +188,54 @@ public PrimitiveConfigDeserializer() {
...
+
+  private static class InterpolatingYamlFactory extends YAMLFactory {
+    @Override
+    protected YAMLParser _createParser(InputStream in, IOContext ctxt) throws IOException {
+      return new InterpolatingYamlParser(ctxt, _getBufferRecycler(), _parserFeatures, _yamlParserFeatures, _objectCodec, _createReader(in, null, ctxt));
+    }
+  }
+
+  private static class InterpolatingYamlParser extends YAMLParser {
+    private static final Pattern SYS_PATTERN = Pattern.compile("\\$\\{sys:([A-Za-z0-9-_.]+)\\}");
```
johnou: I was thinking no static because it's bootstrap only then it may be garbage collected but maybe the jvm is smart enough.

kuujo:
No I just confused myself... it will be

config/src/main/java/io/atomix/core/config/jackson/JacksonConfigProvider.java
kuujo: 🤔

johnou: nope.

Static variables cannot be elected for garbage collection while the class is loaded.

so I guess for it to happen you'd need to enable class unloading.

PR #480, opened 2018-04-07
Miscellaneous static bug analysis fixes.

johnou:


storage/src/main/java/io/atomix/storage/buffer/FileBytes.java
```diff
@@ -249,15 +249,15 @@ public Bytes read(int position, Bytes bytes, int offset, int length) {
     if (bytes.hasArray()) {
       try {
         seekToOffset(position);
-        randomAccessFile.read(bytes.array(), (int) offset, (int) length);
+        randomAccessFile.readFully(bytes.array(), (int) offset, (int) length);
```
johnou: Although RandomAccessFile is not a subclass of InputStream, this method behaves in exactly the same way as the InputStream.read() method of InputStream.
https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#read(byte[],%20int,%20int)
An attempt is made to read as many as len bytes, but a smaller number may be read.

core/src/test/java/io/atomix/core/AtomixTest.java
```diff
@@ -230,6 +231,7 @@ public void onEvent(ClusterEvent event) {
       try {
         queue.put(event);
       } catch (InterruptedException e) {
+        Throwables.propagate(e);
```
kuujo: Isn’t this deprecated now? Not sure what’s supposed to replace it though.

core/src/test/java/io/atomix/core/tree/impl/DocumentTreeTest.java
```diff
@@ -300,7 +300,7 @@ public void testClear() throws Exception {
   /**
    * Tests listeners.
    */
-  @Test
+  @Test(timeout = 45000)
```
kuujo: Can we just create a base class that assigns a timeout to all the tests?

core/src/test/java/io/atomix/core/AtomixTest.java
johnou: Yep, in this cause I would probably just interrupt the thread again, but instead copied code which is like this in about three places..

core/src/test/java/io/atomix/core/tree/impl/DocumentTreeTest.java
johnou: Not to first look, only thing I see is per class at suite level.

core/src/test/java/io/atomix/core/AtomixTest.java
johnou: updated.

core/src/test/java/io/atomix/core/tree/impl/DocumentTreeTest.java
johnou: This guy is the main suspect though I was able to reproduce the never ending test like in Travis.

PR #481, opened 2018-04-09
Simplify cluster configuration

kuujo:
This PR simplifies the cluster configuration format. Rather than requiring two separate sections for core/bootstrap nodes, the `CORE` nodes are just extracted from a single list of nodes for the cluster. To join a consensus cluster, a static set of `CORE` nodes must be defined. To join a primary-backup-based cluster, a single active peer must be defined or multicast bootstrap must be enabled.


core/src/main/java/io/atomix/core/Atomix.java
```diff
@@ -472,23 +471,38 @@ private static ManagedMessagingService buildMessagingService(AtomixConfig config
...
   /**
    * Builds a bootstrap metadata service.
    */
   private static ManagedBootstrapMetadataService buildBootstrapMetadataService(AtomixConfig config) {
-    return new DefaultBootstrapMetadataService(config.getClusterConfig().getBootstrapConfig());
+    boolean hasCoreNodes = config.getClusterConfig().getNodes().stream().anyMatch(node -> node.getType() == Node.Type.CORE);
+    ClusterMetadata metadata = ClusterMetadata.builder()
+        .withNodes(config.getClusterConfig().getNodes()
+            .stream()
+            .filter(node -> !hasCoreNodes || node.getType() == Node.Type.CORE)
```
johnou: Do we really want clients in here too if no core nodes are defined?

kuujo: Hmm... probably not

johnou: the way I read it at least is that if no core nodes are defined client nodes will be considered bootstrap nodes.

kuujo: That is indeed true

kuujo: Fixed

johnou:
@kuujo gitignore core/data/?

core/src/test/java/io/atomix/core/AbstractAtomixTest.java
```diff
@@ -50,36 +51,35 @@ public static void setupAtomix() throws Exception {
...
-    Collection<Node> bootstrapNodes = bootstrapIds.stream()
-        .map(nodeId -> Node.builder(String.valueOf(nodeId))
-            .withType(Node.Type.CORE)
-            .withAddress("localhost", BASE_PORT + nodeId)
-            .build())
+    Collection<Node> nodes = Stream.concat(
+        nodeIds.stream()
+            .map(nodeId -> Node.builder(String.valueOf(nodeId))
+                .withType(Node.Type.CORE)
+                .withAddress("localhost", BASE_PORT + nodeId)
```
johnou: find available port helper method?

johnou:
one more nit and lgtm.

core/src/test/java/io/atomix/core/AbstractAtomixTest.java
kuujo: Added that in #482

PR #490, opened 2018-04-13
Fix various bugs in Atomix agent configuration

kuujo:
This PR fixes various bugs in the agent configuration.

### DIFF
--- a/agent/src/main/java/io/atomix/agent/AtomixAgent.java
+++ b/agent/src/main/java/io/atomix/agent/AtomixAgent.java
@@ -63,50 +63,55 @@ public class AtomixAgent {
     parser.addArgument("node")
         .type(nodeArgumentType.apply(Node.Type.CORE))
         .nargs("?")
-        .metavar("NAME:HOST:PORT")
+        .metavar("NAME@HOST:PORT")
         .required(false)
-        .help("The local node info");
+        .help("The node info for the local node. This should be in the format [NAME@]HOST[:PORT]. " +
+            "If no name is provided, the node name will default to the host. " +
+            "If no port is provided, the port will default to 5679.");
     parser.addArgument("--type", "-t")
         .type(typeArgumentType)
         .metavar("TYPE")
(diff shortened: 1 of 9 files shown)


agent/src/main/java/io/atomix/agent/AtomixAgent.java
```diff
@@ -78,13 +76,13 @@ public static void main(String[] args) throws Exception {
...
         .nargs("*")
-        .type(nodeArgumentType)
+        .type(nodeArgumentType.apply(Node.Type.CORE))
         .metavar("NAME:HOST:PORT")
         .required(false)
         .help("Sets the core cluster configuration");
     parser.addArgument("--bootstrap-nodes", "-b")
         .nargs("*")
-        .type(nodeArgumentType)
+        .type(nodeArgumentType.apply(Node.Type.DATA))
```
johnou: data for bootstrap nodes? I haven't used args4j much..

kuujo: If `--core-nodes` are provided, they'll be used as bootstrap nodes, so the assumption is that any `--bootstrap-nodes` provided will be `DATA` nodes

johnou: what if they aren't? is it corrected with gossip?

kuujo: if `--core-nodes` aren't provided then presumably there _are_ no core nodes. This is an unavoidable limitation of consensus. A static consensus configuration must be provided or correctness is risked.

johnou: I mean what if I start a node as core and configure that node as data in another node.

kuujo: The problem is, if different nodes are started with different sets of `CORE` nodes, they'll count votes differently and can thus elect multiple leaders, leading to split brain. The CORE configuration must be the same on all nodes at startup. There's no way around that limitation. A single source must be assumed - either in a provided configuration or by querying a single external source.

PR #498, opened 2018-04-20
Library updates

johnou:


config/pom.xml
```diff
@@ -47,17 +47,14 @@
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-core</artifactId>
-      <version>${jackson.version}</version>
```
johnou: inherited from dep management in parent pom.

PR #499, opened 2018-04-20
Add operation to query lock status.

tanbangcheng:


johnou:
What's the use case? Given the lock state may change after isLocked has returned I would recommend tryLock then something..

tanbangcheng:
@johnou Just check the state but not lock it, like Redisson's RLock.  
use RLock to mark whether a switch is managed by the Controller (then the Controller locks, and do something),  if not locked, the manager assigns the lock to one of the Controllers

johnou:
@tanbangcheng rebase please?

core/src/main/java/io/atomix/core/lock/impl/DistributedLockService.java
```diff
@@ -49,4 +49,11 @@ default void lock(int lockId) {
   @Command("unlock")
   void unlock(int lockId);
 
+  /**
+   * Query whether the lock state.
+   *
+   * @return {@code true} if this lock is locked, {@code false} otherwise
+   */
+  @Operation(value = "isLocked", type = OperationType.QUERY)
```
johnou: @tanbangcheng query annotation?

johnou:
@tanbangcheng rebase?

PR #503, opened 2018-04-25
Fix keep on restoring.

tanbangcheng:
1. backup index out of order  
```context.protocol().backup(nodeId, request);```
PrimaryRole send backup index 1 ~ 100 and 101 ~ 200 without interval, BackupRole may receive 101 ~ 200 and 1 ~ 100, then request restore.

2. keep on restoring
   PrimaryRole send RestoreResponse at 2300, but in the AsynchronousReplicator queue, there may be operations whose index is less than 2300 waiting to be sent. 
  then RestoreRequest again. 

join backupResponse to fix 1.
clear operations whose index is less than restore index to fix 2.

### DIFF
--- a/protocols/primary-backup/src/main/java/io/atomix/protocols/backup/impl/PrimaryBackupServerContext.java
+++ b/protocols/primary-backup/src/main/java/io/atomix/protocols/backup/impl/PrimaryBackupServerContext.java
@@ -39,6 +39,7 @@ import io.atomix.protocols.backup.protocol.RestoreResponse;
 import io.atomix.protocols.backup.service.impl.PrimaryBackupServiceContext;
 import io.atomix.utils.Managed;
 import io.atomix.utils.concurrent.Futures;
+import io.atomix.utils.concurrent.OrderedFuture;
 import io.atomix.utils.concurrent.ThreadContextFactory;
 
 import java.util.List;
@@ -148,7 +149,16 @@ public class PrimaryBackupServerContext implements Managed<Void> {
           memberGroupService,
           protocol,
           primaryElection);
-      return service.open().thenApply(v -> service);
+
+      OrderedFuture<PrimaryBackupServiceContext> newOrderFuture = new OrderedFuture<>();
+      service.open().whenComplete((v, e) -> {
+        if (e != null) {
+          newOrderFuture.completeExceptionally(e);
+        } else {
+          newOrderFuture.complete(service);
+        }
+      });
+      return newOrderFuture;
     });
   }
 
--- a/protocols/primary-backup/src/main/java/io/atomix/protocols/backup/roles/BackupRole.java
+++ b/protocols/primary-backup/src/main/java/io/atomix/protocols/backup/roles/BackupRole.java
@@ -94,6 +94,9 @@ public class BackupRole extends PrimaryBackupRole {
             applyClose((CloseOperation) operation);
             break;
         }
+
+      } else if (operation.index() < i) {
+        continue;
       } else {
         requestRestore(context.primary());
         break;
--- a/protocols/primary-backup/src/main/java/io/atomix/protocols/backup/service/impl/PrimaryBackupServiceContext.java
+++ b/protocols/primary-backup/src/main/java/io/atomix/protocols/backup/service/impl/PrimaryBackupServiceContext.java
@@ -299,7 +299,6 @@ public class PrimaryBackupServiceContext implements ServiceContext {
     currentOperation = OperationType.COMMAND;
     operationIndex = index;
     currentIndex = index;
-    commitIndex = index;
     currentTimestamp = timestamp;
     service.tick(new WallClockTimestamp(currentTimestamp));
   }
--- a/protocols/primary-backup/src/main/java/io/atomix/protocols/backup/service/impl/PrimaryBackupServiceContext.java
+++ b/protocols/primary-backup/src/main/java/io/atomix/protocols/backup/service/impl/PrimaryBackupServiceContext.java
@@ -288,6 +288,7 @@ public class PrimaryBackupServiceContext implements ServiceContext {
     currentOperation = OperationType.COMMAND;
     operationIndex = index;
     currentIndex = index;
+    commitIndex = index;
     currentTimestamp = timestamp;
   }
 
--- a/protocols/primary-backup/src/main/java/io/atomix/protocols/backup/service/impl/PrimaryBackupServiceContext.java
+++ b/protocols/primary-backup/src/main/java/io/atomix/protocols/backup/service/impl/PrimaryBackupServiceContext.java
@@ -300,6 +300,7 @@ public class PrimaryBackupServiceContext implements ServiceContext {
     operationIndex = index;
     currentIndex = index;
     currentTimestamp = timestamp;
+    setCommitIndex(index);
     service.tick(new WallClockTimestamp(currentTimestamp));
   }
 
--- a/utils/src/main/java/io/atomix/utils/concurrent/OrderedFuture.java
+++ b/utils/src/main/java/io/atomix/utils/concurrent/OrderedFuture.java
@@ -80,6 +80,7 @@ public class OrderedFuture<T> extends CompletableFuture<T> {
           future.completeExceptionally(error);
         }
       }
+      orderedFutures.clear();
     }
   }
 
--- a/utils/src/main/java/io/atomix/utils/concurrent/OrderedFuture.java
+++ b/utils/src/main/java/io/atomix/utils/concurrent/OrderedFuture.java
@@ -33,9 +33,9 @@ import java.util.function.Function;
  */
 public class OrderedFuture<T> extends CompletableFuture<T> {
   private final Queue<CompletableFuture<T>> orderedFutures = new LinkedList<>();
-  private boolean complete;
-  private T result;
-  private Throwable error;
+  private volatile boolean complete;
+  private volatile T result;
+  private volatile Throwable error;
 
   public OrderedFuture() {
     super.whenComplete(this::complete);
@@ -45,18 +45,21 @@ public class OrderedFuture<T> extends CompletableFuture<T> {
    * Adds a new ordered future.
    */
   private CompletableFuture<T> orderedFuture() {
-    synchronized (orderedFutures) {
-      if (complete) {
-        if (error == null) {
-          return CompletableFuture.completedFuture(result);
-        } else {
-          return Futures.exceptionalFuture(error);
+    if (!complete) {
+      synchronized (orderedFutures) {
+        if (!complete) {
+          CompletableFuture<T> future = new CompletableFuture<>();
+          orderedFutures.add(future);
+          return future;
         }
-      } else {
-        CompletableFuture<T> future = new CompletableFuture<>();
-        orderedFutures.add(future);
-        return future;
       }
+    }
+
+    // Completed
+    if (error == null) {
+      return CompletableFuture.completedFuture(result);
+    } else {
+      return Futures.exceptionalFuture(error);
     }
   }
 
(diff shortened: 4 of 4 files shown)

### CODE REVIEW

File: protocols/primary-backup/src/main/java/io/atomix/protocols/backup/roles/AsynchronousReplicator.java [hunk 1]
```diff
@@ -110,7 +118,7 @@ private void backup() {
           context.currentIndex(),
           batch);
       log.trace("Sending {} to {}", request, nodeId);
-      context.protocol().backup(nodeId, request);
+      context.protocol().backup(nodeId, request).join();
```
johnou: Instead of join here can we change the method signature to return completable future and return the cf from backup then propagate that up the call stack?

johnou: at least for add it's quite easy..

```java
  @Override
  public CompletableFuture<Void> replicate(BackupOperation operation) {
    List<CompletableFuture<Void>> futures = new ArrayList<>(context.backups().size());
    for (NodeId backup : context.backups()) {
      futures.add(queues.computeIfAbsent(backup, BackupQueue::new).add(operation));
    }
    context.setCommitIndex(operation.index());
    return Futures.allOf(futures).thenApply(v -> null);
  }
```

protocols/primary-backup/src/main/java/io/atomix/protocols/backup/roles/AsynchronousReplicator.java
```diff
@@ -68,6 +77,7 @@ public void close() {
     private final Queue<BackupOperation> operations = new LinkedList<>();
     private final NodeId nodeId;
     private final Scheduled backupTimer;
+    private CompletableFuture<BackupResponse> lastBackup = CompletableFuture.completedFuture(null);
```
johnou: are add and maybeBackup invoked from the same thread? do we need to sync on backup? volatile?

protocols/primary-backup/src/main/java/io/atomix/protocols/backup/roles/AsynchronousReplicator.java
tanbangcheng: this may make ```replicate``` same as ```SynchronousReplicator```

protocols/primary-backup/src/main/java/io/atomix/protocols/backup/roles/AsynchronousReplicator.java
johnou: I was thinking something like this https://gist.github.com/johnou/55bf838142b38695830f7941368ce9fb ping @kuujo

tanbangcheng: both of them invoked in the same threadContext, maybe I think it doesn't need synchronized

johnou:
tanbangcheng problems with rebase?

tanbangcheng:
@johnou unit test is ok, 
my code work fine after patching this.

johnou:
@tanbangcheng was that the test marked as ignore in the other pull request?

tanbangcheng:
@johnou not, that is not 100% passed, may time out

protocols/primary-backup/src/main/java/io/atomix/protocols/backup/roles/AsynchronousReplicator.java
```diff
@@ -110,8 +123,7 @@ private void backup() {
           context.currentIndex(),
           batch);
       log.trace("Sending {} to {}", request, nodeId);
-      context.protocol().backup(nodeId, request);
-      lastSent = System.currentTimeMillis();
+      return context.protocol().backup(nodeId, request).thenRun(() -> lastSent = System.currentTimeMillis());
```
johnou: Thinking about this more I think it would be safer to drop the backupFuture field and set the timestamp before invoking backup.

tanbangcheng: NettyMessagingService don‘t guarantee message to be sent on order,  if invoke `replicate` too quickly, backupRole may receive  two out-of-order `BackupRequest`, maybe I think need backupFuture to guarantee the order here

johnou: It should keep the order now with our change that collects all the futures.

tanbangcheng: sorry, I don't understand what you mean, 
future of `maybeBackup` is not collected.

johnou: And if we set the timestamp before returning the actual cf it shouldn't be executed in parallel.

tanbangcheng: but we cannot assume the message will be sent within 50ms, 
another case 
*  `maybeBackup` -> `backup` is invoked,  
* 100 `BackupOperation`s in a short time, invoke `backup` again

johnou: @tanbangcheng how about something like this? https://gist.github.com/johnou/55bf838142b38695830f7941368ce9fb

protocols/primary-backup/src/main/java/io/atomix/protocols/backup/roles/BackupRole.java
```diff
@@ -167,6 +167,7 @@ private void requestRestore(NodeId primary) {
 
             context.service().restore(buffer);
             operations.clear();
+            context.setCommitIndex(response.index());
```
johnou: why? isn't it redundant? we set the commit index a few lines above with context.resetIndex

tanbangcheng: I do not know why `resetIndex` not set commitIndex, so i invoke setCommitIndex() here

kuujo:
Needs to be rebased and updated with the final API changes (`NodeId` -> `MemberId`)

protocols/primary-backup/src/main/java/io/atomix/protocols/backup/roles/BackupRole.java
kuujo: That's the only place that method is called, so it should be fine to remove this line and set it inside `resetIndex`

protocols/primary-backup/src/main/java/io/atomix/protocols/backup/roles/AsynchronousReplicator.java
kuujo: It makes sense to me to record the timestamp after the response. Correctness doesn't depend on storing it prior to the request.

kuujo:
I don't like this. This seems to change asynchronous backups to synchronous backups:

> PrimaryRole send backup index 1 ~ 100 and 101 ~ 200 without interval, BackupRole may receive 101 ~ 200 and 1 ~ 100, then request restore.

The underlying protocol is TCP, and the underlying `ClusterMessagingService` is designed to use a single TCP connection for the same type of message, so there's no reason we should need to block replication while a batch is already in flight. Awaiting a response for every batch is a major performance overhead. We should be able to pipeline batches of changes to backups.

protocols/primary-backup/src/main/java/io/atomix/protocols/backup/roles/PrimaryRole.java
```diff
@@ -171,6 +170,9 @@ private ExecuteResponse applyQuery(ExecuteRequest request, Session session) {
     context.service().backup(buffer);
     buffer.flip();
     byte[] bytes = buffer.readBytes(buffer.remaining());
+
+    replicator.removePreviousOperation(request.nodeId(), context.currentIndex());
```
kuujo: This line seems to assume that the primary's response to the backup is going to be received. That seems like it could be dangerous to me. We probably shouldn't remove operations from the primary until it *knows* the backup received and installed the snapshot. Of course, we could just not remove these operations here at all. Instead, the backup can respond to backup requests with the highest index it has received, and the primary can remove indexes it _knows_ the backup already has.

tanbangcheng:
I found that sending two or more Requests before the Service is ready, "thenCompose" will cause out of order.

utils/src/main/java/io/atomix/utils/concurrent/OrderedFuture.java
```diff
@@ -45,19 +45,22 @@ public OrderedFuture() {
    * Adds a new ordered future.
    */
   private CompletableFuture<T> orderedFuture() {
-    synchronized (orderedFutures) {
```
johnou: fields are not volatile and complete, error, result were guarded by this sync, not thread safe anymore?

tanbangcheng: that's my mistake, they should all be volatile

protocols/primary-backup/src/main/java/io/atomix/protocols/backup/impl/PrimaryBackupServerContext.java
```diff
@@ -148,7 +149,16 @@ public Role getRole() {
...
+
+      OrderedFuture<PrimaryBackupServiceContext> newOrderFuture = new OrderedFuture<>();
+      service.open().whenComplete((v, e) -> {
+        if (e != null) {
+          newOrderFuture.completeExceptionally(e);
+        } else {
+          newOrderFuture.complete(service);
+        }
+      });
+      return newOrderFuture;
```
johnou: So what are we chaining while waiting for the service?

johnou:
@kuujo ptal

protocols/primary-backup/src/main/java/io/atomix/protocols/backup/service/impl/PrimaryBackupServiceContext.java
```diff
@@ -288,6 +288,7 @@ public void resetIndex(long index, long timestamp) {
     currentOperation = OperationType.COMMAND;
     operationIndex = index;
     currentIndex = index;
+    commitIndex = index;
```
johnou: not super sure about this.

kuujo: I think @johnou is right. The `commitIndex` should never be decreased, so this should be ensuring that doesn't happen with `setCommitIndex` rather than a simple update.

kuujo:
Actually, I think it's okay for the backup to reset its `commitIndex` when it installs a snapshot, but it also needs to be rejecting the primary's backup requests if it can't successfully apply them (they're not in order or it doesn't have a snapshot) so the primary doesn't consider a change to have been replicated when the backup actually needs a restore.

kuujo:
Sorry, I can't make up my mind. I've gone back and forth, but @johnou is still right. There's no legitimate reason for the backup to revert its `commitIndex` since it should only be committing changes that it already has. If it's not then that is likely a bug elsewhere.

protocols/primary-backup/src/main/java/io/atomix/protocols/backup/service/impl/PrimaryBackupServiceContext.java
```diff
@@ -299,7 +299,6 @@ public void resetIndex(long index, long timestamp) {
     currentOperation = OperationType.COMMAND;
     operationIndex = index;
     currentIndex = index;
-    commitIndex = index;
```
kuujo: I think this just need to be updated, it just can't be decreased.

johnou: ohh you mean like this? commitIndex = Math.max(commitIndex, index);

kuujo: yeah... or `setCommitIndex` has the `max` call in it already

PR #505, opened 2018-04-25
Backup OOO

johnou:
Follow up of https://github.com/atomix/atomix/pull/503


tanbangcheng:
`backup` clear and send all `operations`, not 100 operations per backup,  maybe i think should not chain backups

johnou:
@tanbangcheng I know, the point was to maintain ordering but not drop any backups.

kuujo:
Needs to be rebased and updated with the latest API changes.

PR #508, opened 2018-04-26
Add node tags for generic classification of nodes

kuujo:
This PR adds support for assigning arbitrary tags to nodes for use in member groups and for other general information about nodes.

```java
Node node = Node.builder("foo")
  .withAddress("localhost:5000")
  .withTags("foo", "bar")
  .build();
```

### DIFF
--- a/cluster/src/main/java/io/atomix/cluster/Node.java
+++ b/cluster/src/main/java/io/atomix/cluster/Node.java
@@ -15,9 +15,12 @@
  */
 package io.atomix.cluster;
 
+import com.google.common.collect.ImmutableSet;
 import io.atomix.utils.net.Address;
 
+import java.util.HashSet;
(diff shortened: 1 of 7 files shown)


cluster/src/main/java/io/atomix/cluster/Node.java
```diff
@@ -360,12 +375,36 @@ public Builder withHost(String host) {
...
 
+    /**
+     * Sets the node tags.
+     *
+     * @param tags the node tags
+     * @return the node builder
+     * @throws NullPointerException if the tags are null
+     */
+    public Builder withTags(Set<String> tags) {
+      this.tags = checkNotNull(tags, "tags cannot be null");
```
johnou: iterate tags and invoke addTag to validate the contents of the set?

PR #518, opened 2018-04-27
Static code analysis fixes

johnou:


config/pom.xml
```diff
@@ -76,7 +76,7 @@
         <configuration>
           <instructions>
             <Export-Package>
-              io.atomix.core.config.jackson.*
```
johnou: or the meta-inf/services fails

johnou: (compile error in intellij idea)

PR #519, opened 2018-04-27
Ensure system partition group is closed

kuujo:
This PR fixes a leak when shutting down `Atomix` instances. Currently, the system partition group may not be shutdown. Additionally, primary-backup `PrimaryRole` threads were not being stopped when primary-backup partitions were shutdown.


protocols/primary-backup/src/main/java/io/atomix/protocols/backup/service/impl/PrimaryBackupServiceContext.java
```diff
@@ -577,6 +577,7 @@ private void changeRole(PrimaryTerm term) {
   public CompletableFuture<Void> close() {
     clusterMembershipService.removeListener(membershipEventListener);
     primaryElection.removeListener(primaryElectionListener);
+    this.role.close();
```
kuujo: remove `this`

PR #521, opened 2018-04-27
Change transaction service partition

kuujo:
The `TransactionService` is currently using the default partition group for transaction state. This should be done in the system partition. This also removes the default partition group and defaults the partition group only if a single group is provided. This is because groups are no longer ordered since they can be discovered from other nodes.

### DIFF
--- a/primitive/src/main/java/io/atomix/primitive/DistributedPrimitiveBuilder.java
+++ b/primitive/src/main/java/io/atomix/primitive/DistributedPrimitiveBuilder.java
@@ -15,6 +15,7 @@
  */
 package io.atomix.primitive;
 
+import com.google.common.base.Joiner;
 import io.atomix.primitive.partition.PartitionGroup;
 import io.atomix.primitive.protocol.PrimitiveProtocol;
 import io.atomix.primitive.protocol.PrimitiveProtocolConfig;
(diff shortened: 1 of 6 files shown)


kuujo:
Now let's see how the tests do :-)

kuujo:
Transaction tests still seem to be failing sometimes.

primitive/src/main/java/io/atomix/primitive/DistributedPrimitiveBuilder.java
```diff
@@ -159,7 +162,12 @@ public PrimitiveProtocol protocol() {
     if (protocol == null) {
       PrimitiveProtocolConfig protocolConfig = config.getProtocolConfig();
       if (protocolConfig == null) {
-        protocol = managementService.getPartitionService().getDefaultPartitionGroup().newProtocol();
+        Collection<PartitionGroup> partitionGroups = managementService.getPartitionService().getPartitionGroups();
+        if (partitionGroups.size() == 1) {
+          protocol = partitionGroups.iterator().next().newProtocol();
+        } else {
+          throw new ConfigurationException("Primitive protocol is ambiguous: " + partitionGroups.size() + " partition groups found");
```
johnou: could we print the types here?

PR #527, opened 2018-05-02
Implement caching for ConsistentMultimap primitive

kuujo:
This PR adds a cached implementation of `ConsistentMultimap`


core/src/main/java/io/atomix/core/multimap/impl/CachingAsyncConsistentMultimap.java
```diff
@@ -0,0 +1,180 @@
...
+
+  @Override
+  public CompletableFuture<Boolean> containsKey(K key) {
+    return get(key).thenApply(value -> value != null && !value.value().isEmpty());
+  }
+
+  @Override
+  public CompletableFuture<Boolean> put(K key, V value) {
+    return super.put(key, value)
+        .whenComplete((r, e) -> cache.invalidate(key));
```
johnou: invalidate on put, or update?

kuujo: Both. The cache can have a future that returns `null` from a prior read on an empty key.

PR #528, opened 2018-05-02
Support for primitive service configurations

kuujo:
This PR adds support for primitive service configurations. This is needed to be able to add configurations that need to be represented in primitive services. For example, `ConsistentTreeMap` sort order, `ConsistentMultimap` collection type, or `DistributedSemaphore` permit count. Service configuration classes can be defined by the `PrimitiveType` and are provided to the primitive service factory method.

### DIFF
--- a/core/src/main/java/io/atomix/core/queue/WorkQueueType.java
+++ b/core/src/main/java/io/atomix/core/queue/WorkQueueType.java
@@ -30,7 +30,7 @@ import static com.google.common.base.MoreObjects.toStringHelper;
  * Work queue primitive type.
  */
 public class WorkQueueType<E> implements PrimitiveType<WorkQueueBuilder<E>, WorkQueueConfig, WorkQueue<E>, ServiceConfig> {
-  private static final String NAME = "WORK_QUEUE";
+  private static final String NAME = "work-queue";
 
   /**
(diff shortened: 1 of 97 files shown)


core/src/main/java/io/atomix/core/queue/WorkQueueType.java
```diff
@@ -22,16 +22,14 @@
...
+import io.atomix.primitive.service.ServiceConfig;
 
 import static com.google.common.base.MoreObjects.toStringHelper;
 
 /**
  * Work queue primitive type.
  */
-public class WorkQueueType<E> implements PrimitiveType<WorkQueueBuilder<E>, WorkQueueConfig, WorkQueue<E>> {
+public class WorkQueueType<E> implements PrimitiveType<WorkQueueBuilder<E>, WorkQueueConfig, WorkQueue<E>, ServiceConfig> {
   private static final String NAME = "WORK_QUEUE";
```
johnou: lowercase to match change for map?

johnou: lowercase to match change for map?

kuujo: Ahh yes I meant to change all of them

PR #530, opened 2018-05-02
Member's metadata

### LINKED ISSUE #529: Member's metadata
In atomix v1, a member had an option to associate some meta-data when joining a cluster, now I see that a member has tags but it is less flexible that meta-data, would it possible to switch from tags to meta-data or have a way to add arbitrary meta-data to a member when joining a cluster ?

lburgazzoli:
Fixes #529

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ core/data/
 agent/data/
 data/
 .data/
+.sts4-cache/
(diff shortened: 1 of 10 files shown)


johnou:
@lburgazzoli simple unit test to show it replicates?

lburgazzoli:
@johnou can you point me to a simple test doing something similar ?

johnou:
@lburgazzoli something like io.atomix.core.AtomixTest#testClientJoinLeavePersistent but checking tags are the same between nodes instead of events, remember it's a hashmap so order is not deterministic.

lburgazzoli:
done

johnou:
@lburgazzoli thanks! :rocket:

PR #533, opened 2018-05-02
Distributed primitive proxies

kuujo:
This PR refactors the client-side primitive APIs to support the use of Java proxies for primitive state machines. The proxies work by providing a proxy interface for the primitive service to implement, e.g.:

```java
public interface ConsistentMapService {
  @Operation(value = "containsKey", type = OperationType.QUERY)
  boolean containsKey(String key);

  @Operation(value = "containsValue", type = OperationType.QUERY)
  boolean containsValue(byte[] value);

  @Operation(value = "get", type = OperationType.QUERY)
  Versioned<byte[]> get(String key);

  @Operation(value = "put", type = OperationType.COMMAND)
  Versioned<byte[]> put(String key, byte[] value);
}
```

The proxy interface must then be implemented by the service implementation:

```java
public class ConsistentMapServiceImpl extends AbstractPrimitiveService implements ConsistentMapService {
  ...
}
```

Then, the service interface can be used as a proxy in the asynchronous primitive client:

```java
public class ConsistentMapProxy extends AbstractAsyncPrimitiveProxy<AsyncConsistentMap<String, byte[]>, ConsistentMapProxy> implements AsyncConsistentMap<String, byte[]> {
  public ConsistentMapProxy(PrimitiveProxy proxy, PrimitiveRegistry registry) {
    super(ConsistentMapService.class, proxy, registry);
  }

  @Override
  public CompletableFuture<Boolean> containsKey(String key) {
    return applyBy(key, partition -> partition.containsKey(key));
  }

  @Override
  public CompletableFuture<Boolean> containsValue(byte[] value) {
    return applyAll(partition -> partition.containsValue(value))
      .thenApply(results -> results.filter(Predicate.isEqual(true)).findFirst().orElse(false));
  }

  @Override
  public CompletableFuture<Versioned<byte[]>> put(String key, byte[] value) {
    return applyBy(key, partition -> partition.put(key, value));
  }
}
```

The `acceptAll`/`acceptOn`/`acceptBy` and `applyAll`/`applyOn`/`applyBy` methods provide a service proxy to a `Consumer` or `Function` for operating on all partitions or on a specific partition. The method call is then converted into a `PrimitiveOperation` to execute on the partitions. So, `applyBy(key, partition -> partition.containsKey(key))` essentially invokes the `containsKey` method on the service in the partition to which the given key maps, and `applyAll(partition -> partition.containsValue(value))` invokes the `containsValue` method on the services in every partition, aggregating the results into a `Stream`.

### DIFF
--- a/primitive/src/main/java/io/atomix/primitive/impl/AbstractAsyncPrimitiveProxy.java
+++ b/primitive/src/main/java/io/atomix/primitive/impl/AbstractAsyncPrimitiveProxy.java
@@ -1,0 +1,239 @@
+/*
+ * Copyright 2018-present Open Networking Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.atomix.primitive.impl;
+
+import com.google.common.base.Defaults;
+import com.google.common.collect.Maps;
+import io.atomix.primitive.AsyncPrimitive;
+import io.atomix.primitive.PrimitiveException;
+import io.atomix.primitive.PrimitiveRegistry;
+import io.atomix.primitive.operation.OperationId;
+import io.atomix.primitive.operation.Operations;
+import io.atomix.primitive.operation.PrimitiveOperation;
+import io.atomix.primitive.partition.PartitionId;
+import io.atomix.primitive.proxy.PartitionProxy;
+import io.atomix.primitive.proxy.PrimitiveProxy;
+import io.atomix.utils.concurrent.Futures;
+
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.util.Collection;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+/**
+ * Abstract asynchronous primitive that provides proxies.
+ */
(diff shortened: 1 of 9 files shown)


primitive/src/main/java/io/atomix/primitive/impl/AbstractAsyncPrimitiveProxy.java
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  /**
+   * Service proxy invocation handler.
+   * <p>
+   * The invocation handler
+   */
+  private class ServiceProxyHandler implements InvocationHandler {
+    private final PartitionProxy proxy;
+    private final ThreadLocal<CompletableFuture> future = new ThreadLocal<>();
```
johnou: this screams trouble to me, are we super sure the future is completed / mutated always on the right thread?

kuujo: The calling thread is getting the future. It's just a trick to get a future from a proxy object. Yes, the future will always be read from the `ThreadLocal` immediately after the proxy is used.

kuujo: The alternative would be using ByteBuddy to generate a class with a different return type (`CompletableFuture`) so a synchronous method call on the proxy object can return a future, but that's a lot of overhead for something that can be done this way. The `ThreadLocal` is just to ensure multiple threads can use the proxy at the same time.

primitive/src/main/java/io/atomix/primitive/impl/AbstractAsyncPrimitiveProxy.java
```diff
@@ -0,0 +1,239 @@
...
+
+    /**
+     * Invokes a void method on the underlying proxy.
+     *
+     * @param operation the operation to perform on the proxy
+     * @return the resulting void future
+     */
+    CompletableFuture<Void> accept(Consumer<S> operation) {
+      operation.accept(proxy);
+      return handler.getResultFuture();
```
kuujo: @johnou this is an example of the proxy being used and then the future being immediately read from it.

primitive/src/main/java/io/atomix/primitive/impl/AbstractAsyncPrimitiveProxy.java
kuujo: This can also be done another way - by creating thread local proxies rather than storing thread local futures, but it's probably a distinction without much of a difference.

PR #537, opened 2018-05-03
Add partition group membership service to track owning members

kuujo:
This PR adds a `PartitionGroupMembershipService` to be used by the `HashBasedPrimaryElection` to ensure only nodes participating in a partition group participate in eventually consistent primary elections.


kuujo:
I think the `HashBasedPrimaryElection` may need to ensure terms are updated when the distributed counter is updated 🤔

PR #547, opened 2018-05-09
Bootstrap both the sender and receiver during startup of partition group service

kuujo:
When nodes are started sequentially, the eventually consistent primary election may not eventually resolve to the correct leader because of the way partition group membership is propagated at startup. This change sends the partition group information to each peer at startup and receives those peers' partition group information back as well.


kuujo:
Took a few tries, but tests passed and are getting there.

PR #549, opened 2018-05-10
Ensure persistent members remain removed from the cluster once shutdown

kuujo:
This PR fixes a bug in removing persistent members from the cluster. After a persistent member is removed, it may actually be re-added by a heartbeat attempt. This simply checks whether a persistent member is present in the metadata before adding it.

### DIFF
--- a/cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMembershipService.java
+++ b/cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMembershipService.java
@@ -262,6 +262,10 @@ public class DefaultClusterMembershipService
    * Activates the given node.
    */
   private void activateNode(Member member) {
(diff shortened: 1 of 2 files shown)


cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMembershipService.java
```diff
@@ -262,6 +262,10 @@ private void handleBroadcastMessage(byte[] message) {
    * Activates the given node.
    */
   private void activateNode(Member member) {
+    if (member.type() == Member.Type.PERSISTENT && !persistentMetadataService.getMetadata().nodes().contains(member)) {
```
johnou: Rename nodes to members?

kuujo: Ahh yeah we should do that separately

PR #555, opened 2018-05-11
Fix missing quotes in JavaDoc headers

kuujo:


kuujo:
Really should remove the `@author` tags altogether, but that's a separate issue

PR #556, opened 2018-05-11
Ensure revisions are correctly restored from snapshots

kuujo:
This PR fixes a bug with restoring primitive services from snapshots. It correctly populates existing services with previous/next revision information for propagation.


kuujo:
Actually, if we're basing revision order only on the fact that a revision number is unique, they should be ordered by the service ID rather than by the revision number.

PR #557, opened 2018-05-17
Clean up primitive REST APIs

kuujo:
This PR cleans up various aspects of REST APIs for primitives. The most significant change is updating the semaphore API to make it consistent with other primitives. Recommend reviewing one commit at a time. Improvements to the REST API are ongoing.


rest/src/test/java/io/atomix/rest/impl/VertxRestServiceTest.java
```diff
@@ -174,6 +174,156 @@ public void testMessages() throws Exception {
...
+        .then()
+        .body(equalTo("1"));
+  }
+
+  @Test
+  public void testValue() throws Exception {
+    JsonNodeFactory jsonFactory = JsonNodeFactory.withExactBigDecimals(true);
+    JsonNode json = jsonFactory.objectNode()
+        .put("type", "value")
+        .set("protocol", jsonFactory.objectNode()
```
johnou: should be protocolConfig right?

kuujo: Should be `protocol`. Shouldn't have removed the property naming strategy

PR #565, opened 2018-05-22
Support large event batches to Raft clients

kuujo:
This PR modifies the Raft session implementation to support streaming events to clients in batches. The batching is done transparently after events are sent at a given index.


kuujo:
This change isn't currently completely efficient. It still holds all events for a given index in memory until they're acked. It could allow the individual batches to be acked/resent so all the events for an index don't have to be held.

PR #566, opened 2018-05-23
Add com.sun.jmx.remote.internal package to utils bundle imports

kuujo:
Add a missing package to the felix configuration


johnou:
Wth are we using com.sun.jmx.remote.internal for? :D

kuujo:
No idea where this comes from, but I ran into it when upgrading Atomix in ONOS.

PR #567, opened 2018-05-23
Replicate member metadata changes in ClusterMembershipService

kuujo:
This PR makes the `Member#metadata()` map modifiable and adds an event to notify listeners of changes to the metadata. This can be used to replicate state changes associated with nodes.

### DIFF
--- a/cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMembershipService.java
+++ b/cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMembershipService.java
@@ -246,20 +246,13 @@ public class DefaultClusterMembershipService
   /**
    * Activates the given member.
    */
-  private void activateMember(Member member) {
+  private void activateMember(StatefulMember member) {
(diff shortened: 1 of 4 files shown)


cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMembershipService.java
```diff
@@ -270,6 +263,11 @@ private void activateMember(Member member) {
       LOGGER.info("{} - Member activated: {}", localMember.id(), existingMember);
       existingMember.setState(State.ACTIVE);
       post(new ClusterMembershipEvent(ClusterMembershipEvent.Type.MEMBER_ADDED, existingMember));
+    } else if (!existingMember.metadata().equals(member.metadata())) {
```
johnou: I'm pretty sure this is won't work.

johnou: Beg your pardon, Map<String, String>, should be fine.

PR #580, opened 2018-05-31
Do not override local member metadata when replicated

kuujo:
Fixes a race condition in member metadata replication. The local member should be the source of truth for its own metadata and not accept metadata changes for itself.

### DIFF
--- a/cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMembershipService.java
+++ b/cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMembershipService.java
@@ -91,6 +91,7 @@ public class DefaultClusterMembershipService
 
   private final AtomicBoolean started = new AtomicBoolean();
   private final StatefulMember localMember;
+  private volatile Map<String, String> localMetadata;
   private final Map<MemberId, StatefulMember> members = Maps.newConcurrentMap();
   private final Map<MemberId, PhiAccrualFailureDetector> failureDetectors = Maps.newConcurrentMap();
   private final Consumer<byte[]> broadcastListener = this::handleBroadcastMessage;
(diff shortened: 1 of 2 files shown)


cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMembershipService.java
```diff
@@ -259,11 +259,11 @@ private void activateMember(StatefulMember member) {
           localMember.rack(),
           localMember.host(),
           localMember.metadata())));
-    } else if (existingMember.getState() == State.INACTIVE) {
+    } else if (!existingMember.id().equals(localMember.id()) && existingMember.getState() == State.INACTIVE) {
```
johnou: shouldn't we still post but not update the local map?

kuujo: This code is only hit when a remote node sends a message to the local node, in which case there could be a race (the remote node is sending old information about the local node). But what's actually needed is a separate check for the local node to ensure it fires an event.

kuujo: Never mind. This patch is all wrong. An event for the _local_ member's metadata is all that's needed.

cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMembershipService.java
```diff
@@ -162,13 +163,19 @@ private void handleBroadcastMessage(byte[] message) {
   /**
    * Sends heartbeats to all peers.
    */
-  private CompletableFuture<Void> sendHeartbeats() {
+  private synchronized CompletableFuture<Void> sendHeartbeats() {
+    Member localMember = getLocalMember();
+    if (!localMember.metadata().equals(localMetadata)) {
```
johnou: wouldn't double checked locking make more sense here?

johnou:
```java
  private CompletableFuture<Void> sendHeartbeats() {
    Member localMember = getLocalMember();
    if (!localMember.metadata().equals(localMetadata)) {
      sync(this) {
        if (!localMember.metadata().equals(localMetadata)) {
          localMetadata = localMember.metadata();
          post(new ClusterMembershipEvent(ClusterMembershipEvent.Type.MEMBER_UPDATED, localMember));
        }
      }
    }
```

kuujo: I was just being lazy :-P

johnou: Less sync ftw.

kuujo: Now I have to fix it because you called me out...

PR #592, opened 2018-06-06
Rename event service getter for consistency with service name

kuujo:


cluster/src/main/java/io/atomix/cluster/AtomixCluster.java
```diff
@@ -189,7 +189,7 @@ public ClusterCommunicationService getCommunicationService() {
    *
    * @return the cluster event service
    */
-  public ClusterEventService getEventingService() {
+  public ClusterEventService getEventService() {
     return eventingService;
```
johnou: variable too?

kuujo: oh yeah

PR #593, opened 2018-06-06
Update ConsistentMultimap implementation to use service proxies

kuujo:


core/src/main/java/io/atomix/core/multimap/ConsistentMultimapType.java
```diff
@@ -48,9 +55,20 @@ public String name() {
     return NAME;
   }
 
+  @Override
+  public Namespace namespace() {
```
johnou: HashMultiset.create()
ImmutableSet.of()
ImmutableList

classes defined in super namespace?

core/src/main/java/io/atomix/core/multimap/impl/ConsistentSetMultimapProxy.java
```diff
@@ -26,180 +26,140 @@
...
-    return invokeBy(key, GET, new Get(key));
+    return applyBy(key, service -> service.get(key));
   }
 
   @Override
   public CompletableFuture<Set<String>> keySet() {
-    return this.<Set<String>>invokeAll(KEY_SET)
-        .thenApply(results -> results.reduce((s1, s2) -> ImmutableSet.<String>builder().addAll(s1).addAll(s2).build()).orElse(ImmutableSet.of()));
+    return applyAll(service -> service.keySet())
+        .thenApply(results -> results.reduce((s1, s2) ->
```
johnou: wouldn't collect make more sense here?

core/src/main/java/io/atomix/core/multimap/impl/ConsistentSetMultimapService.java
```diff
@@ -1,723 +1,228 @@
...
+   * @return true if the map changes because of this call, false otherwise.
    */
-  protected Multiset<String> keys(Commit<Void> commit) {
-    Multiset keys = HashMultiset.create();
-    backingMap.forEach((key, mapEntryValue) -> {
-      keys.add(key, mapEntryValue.values().size());
-    });
-    return keys;
-  }
+  @Command("removeAllValues")
```
johnou: do we check for conflicting command annotations?

kuujo: yep

core/src/main/java/io/atomix/core/multimap/ConsistentMultimapType.java
kuujo: yep

core/src/main/java/io/atomix/core/multimap/impl/ConsistentSetMultimapProxy.java
kuujo: `results` is a `Stream<Set<String>>` so it has to be reduced before it can be collected

johnou: Gotcha

kuujo:
Fix the other issues

PR #595, opened 2018-06-07
Update WorkQueue implementation to use service proxies

kuujo:


core/src/main/java/io/atomix/core/queue/impl/DefaultWorkQueueService.java
```diff
@@ -0,0 +1,228 @@
...
+  @Override
+  public void unregister() {
+    registeredWorkers.remove(getCurrentSession().sessionId());
+  }
+
+  @Override
+  public void add(Collection<byte[]> items) {
+    AtomicInteger itemIndex = new AtomicInteger(0);
+    items.forEach(item -> {
+      String taskId = String.format("%d:%d:%d",
```
johnou: Might be cheaper to create a simple key class and use that instead of concatenating three strings.

kuujo: I've also found `String.format` to be incredible costly. It compiles a regex on every call

kuujo: Unfortunately, I probably can't break the API in this case.

PR #612, opened 2018-06-09
Restore allocated heap buffer after it is used.

johnou:


protocols/primary-backup/src/main/java/io/atomix/protocols/backup/roles/PrimaryRole.java
```diff
@@ -160,19 +160,23 @@ private ExecuteResponse applyQuery(ExecuteRequest request, Session session) {
...
+      Collection<PrimaryBackupSession> sessions = context.getSessions();
+      buffer.writeInt(sessions.size());
+      for (Session session : sessions) {
+        buffer.writeLong(session.sessionId().id());
+        buffer.writeString(session.memberId().id());
+      }
+
+      context.service().backup(new DefaultBackupOutput(buffer, context.service().serializer()));
+      buffer.flip();
+      byte[] bytes = buffer.readBytes(buffer.remaining());
```
johnou: readBytes creates a new byte array so it's safe to dispose of the bytebuffer.

PR #615, opened 2018-06-09
Ensure primary-backup roles are closed in the correct thread

kuujo:
Chance for `ConcurrentModificationException` when shutting down a primary-backup server. The roles should be shutdown in the server thread.


protocols/primary-backup/src/main/java/io/atomix/protocols/backup/service/impl/PrimaryBackupServiceContext.java
```diff
@@ -621,9 +621,13 @@ private void changeRole(PrimaryTerm term) {
...
-    clusterMembershipService.removeListener(membershipEventListener);
-    primaryElection.removeListener(primaryElectionListener);
-    role.close();
-    return CompletableFuture.completedFuture(null);
+    CompletableFuture<Void> future = new CompletableFuture<>();
+    threadContext.execute(() -> {
+      clusterMembershipService.removeListener(membershipEventListener);
+      primaryElection.removeListener(primaryElectionListener);
+      role.close();
+      future.complete(null);
```
johnou: I would consider wrapping completing the future in a finally block to ensure it's always set, imagine close of role is refactored and then throws an exception..

kuujo: true

johnou:
lgtm

PR #622, opened 2018-06-10
Check group type names when comparing replicated partition group types

kuujo:
The current check uses `Object#equals` which compares identities and results in exceptions during partition group replication.


johnou:
@kuujo what exception(s)? Can we also improve the equals implementation?

PR #623, opened 2018-06-11
Do not cancel Netty timeout future when connection is lost

kuujo:
During shutdown, when a connection is lost to another node, a node may cancel the future used to time out messages. This leads to some messages never being timed out which ultimately causes a node to block indefinitely when being shut down. This future should only be cancelled when the `NettyMessagingService` itself is being shut down.


johnou:
Changes already in master.

PR #624, opened 2018-06-11
Ensure primary-backup partition group shuts down thread factory

kuujo:
Minor cleanup of shutdown for primary-backup partition groups.


protocols/primary-backup/src/main/java/io/atomix/protocols/backup/partition/PrimaryBackupPartitionGroup.java
```diff
@@ -182,7 +182,7 @@ public PrimaryBackupPartition getPartition(PartitionId partitionId) {
     List<CompletableFuture<Void>> futures = partitions.values().stream()
         .map(PrimaryBackupPartition::close)
         .collect(Collectors.toList());
-    return CompletableFuture.allOf(futures.toArray(new CompletableFuture[futures.size()])).thenRun(() -> {
+    return CompletableFuture.allOf(futures.toArray(new CompletableFuture[futures.size()])).whenCompleteAsync((r, e) -> {
```
johnou: Using common fjp on purpose?

johnou: probably worth a comment explaining why we are using common fjp.

PR #629, opened 2018-06-11
Simplify Netty shutdown.

johnou:


cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyMessagingService.java
```diff
@@ -634,13 +634,14 @@ private Bootstrap bootstrapClient(Address address) {
   @Override
   public CompletableFuture<Void> stop() {
     if (started.compareAndSet(true, false)) {
-      CompletableFuture<Void> future = new CompletableFuture<>();
-      serverChannel.close().addListener(f -> {
-        future.complete(null);
-      });
-      return future.whenCompleteAsync((result, error) -> {
+      return CompletableFuture.supplyAsync(() -> {
```
johnou: @kuujo running this on a fjp common thread isn't strictly necessary but given shutdown graceful can block for 15 seconds might be nicer instead of doing the work on the thread context thread?

PR #630, opened 2018-06-11
Ensure primary-backup client/server threads are cleaned up

kuujo:


protocols/primary-backup/src/main/java/io/atomix/protocols/backup/service/impl/PrimaryBackupServiceContext.java
```diff
@@ -631,6 +631,6 @@ private void changeRole(PrimaryTerm term) {
         future.complete(null);
       }
     });
-    return future;
+    return future.thenRunAsync(() -> threadContext.close());
```
johnou: isn't this closed in PrimaryBackupServerContext..?

johnou: oh context, not the factory.

kuujo: yeah it's redundant

kuujo: Actually, no it's not. If a partition shuts down a server without shutting down the `ThreadContextFactory` then the thread context should still be shutdown here.

PR #639, opened 2018-06-13
Add distributed set tests and ensure set primitive type is correctly assigned

kuujo:
There was a bug in `DistributedSet` due to the underlying `ConsistentMap` not using the correct serializer. This ensures the underlying map is constructed with the correct serializer and `DistributedSet` returns the `DistributedSetType` as its type. Also adds tests.


core/src/main/java/io/atomix/core/set/impl/DelegatingDistributedSetBuilder.java
```diff
@@ -47,7 +48,7 @@ public DelegatingDistributedSetBuilder(String name, DistributedSetConfig config,
   public CompletableFuture<DistributedSet<E>> buildAsync() {
     ProxyClient<ConsistentMapService> proxy = protocol().newProxy(
         name(),
-        primitiveType(),
```
johnou: is this really necessary?

PR #660, opened 2018-06-26
Run tests against multiple JDKs.

johnou:


johnou:
Relies on https://github.com/atomix/atomix/pull/664

kuujo:
We should probably ensure it's being run on Oracle and OpenJDK still too

PR #663, opened 2018-06-27
Implement minimum/maximum thread counts for various thread pools

kuujo:
This PR implements minimum and maximum default thread counts for various thread pools used in Raft/primary-backup protocols.

### DIFF
--- a/core/src/main/java/io/atomix/core/Atomix.java
+++ b/core/src/main/java/io/atomix/core/Atomix.java
@@ -252,7 +252,7 @@ public class Atomix extends AtomixCluster implements PrimitivesService {
     super(config.getClusterConfig());
     config.getProfiles().forEach(profile -> registry.profiles().getProfile(profile).configure(config));
     this.executorService = Executors.newScheduledThreadPool(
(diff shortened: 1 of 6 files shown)


kuujo:
Some follow up work needs to be done to ensure Raft partition clients/servers in the same partition group are sharing a thread pool just as the primary-backup clients/servers share a thread pool within a partition group. Otherwise the number of threads can grow too large with enough Raft partitions.

core/src/main/java/io/atomix/core/Atomix.java
```diff
@@ -252,7 +252,7 @@ private Atomix(AtomixConfig config, AtomixRegistry registry) {
     super(config.getClusterConfig());
     config.getProfiles().forEach(profile -> registry.profiles().getProfile(profile).configure(config));
     this.executorService = Executors.newScheduledThreadPool(
-        Runtime.getRuntime().availableProcessors(),
+        Math.max(Math.min(Runtime.getRuntime().availableProcessors() * 2, 8), 4),
```
johnou: so between 4-8?

Issue #644, opened 2018-06-22
How to control memory usage of log entry when StorageLevel is MAPPED


zhaozhongguo:
Hi:

When the raft StorageLevel is set to MAPPED mode, if the load is always at a high level, the log compact will not be executed, and the memory will continue to grow. Should you provide an interface that configures memory usage like isRunningOutOfDiskSpace

johnou:
@zhaozhongguo would you be willing to provide a pull request?

johnou:
Thanks!

zhaozhongguo:
welcome 👍

PR #672, opened 2018-06-28
 provider memory check interface for storage

zhaozhongguo:
issue #644 How to control memory usage of log entry when StorageLevel is MAPPED 

When the raft StorageLevel is set to MAPPED mode, if the load is always at a high level, the log compact will not be executed, and the memory will continue to grow.


johnou:
Instead of introducing a new interface, can we just use io.atomix.storage.statistics.StorageStatistics#getFreeSpace?

storage/src/main/java/io/atomix/storage/statistics/StorageStatistics.java
```diff
@@ -16,15 +16,19 @@
 package io.atomix.storage.statistics;
 
 import java.io.File;
+import java.lang.management.ManagementFactory;
+import com.sun.management.OperatingSystemMXBean;
```
johnou: can we use the java.lang.management import for OperatingSystemMXBean instead?

johnou: can we use the java.lang.management import for OperatingSystemMXBean instead?

johnou: OperatingSystemMXBean operatingSystemMXBean = ManagementFactory.getOperatingSystemMXBean();

zhaozhongguo: there is no memory api in java.lang.management.OperatingSystemMXBean (getFreePhysicalMemorySize getTotalPhysicalMemorySize)

zhaozhongguo: only provided in com.sun.management

johnou: could we use Runtime.getRuntime().freeMemory() and Runtime.getRuntime().totalMemory() instead?

zhaozhongguo: Runtime.getRuntime()  can only get  memory info about jvm process , not host

zhaozhongguo: seems only com.sun.management.OperatingSystemMXBean or provide a interface to user

johnou: doesn't the jvm memory size make more sense? we could also instanceof the mxbean to make sure there is not a classcastexception, and if it's null, then fallback to runtime? @kuujo wdyt?

zhaozhongguo: In our scenario, the host memory keeps growing and jvm's memory cannot represent the host's memory

zhaozhongguo: The old API in StorageStatistics is also to get the host disk space. Maybe we don't need to think about jvm. Are there any other better suggestions?

johnou: let's wait to hear from @kuujo

zhaozhongguo: ok

johnou: mbean server would be another possibility 

```java
MBeanServer mBeanServer = ManagementFactory.getPlatformMBeanServer();
    Object attribute = mBeanServer.getAttribute(new ObjectName("java.lang","type","OperatingSystem"), "TotalPhysicalMemorySize");
```

johnou: @zhaozhongguo I spoke with kuujo, please go ahead with the mbean approach.

storage/src/main/java/io/atomix/storage/statistics/StorageStatistics.java
```diff
@@ -53,4 +64,32 @@ public long getFreeSpace() {
...
+   * Returns the amount of free memory remaining.
+   *
+   * @return the amount of free memory remaining
+ * @throws MBeanException
+ * @throws ReflectionException
+ * @throws MalformedObjectNameException
+ * @throws AttributeNotFoundException
+ * @throws InstanceNotFoundException
+   */
+  public long getFreeMemory() throws InstanceNotFoundException, AttributeNotFoundException, MalformedObjectNameException, ReflectionException, MBeanException {
```
johnou: instead of throwing five types of checked exceptions I think it would make sense to catch in here and return -1 on failure wdyt?

zhaozhongguo: we cannot print log here, people will not get the proper exception reason . wdyt?

johnou: Create the object name objects in the constructor?

protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
```diff
@@ -64,6 +64,12 @@
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 
+import javax.management.AttributeNotFoundException;
+import javax.management.InstanceNotFoundException;
```
johnou: We can remove these imports now right?

protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftServiceManager.java
```diff
@@ -160,10 +186,13 @@ private void scheduleSnapshots() {
       // Determine whether the node is running out of disk space.
       boolean runningOutOfDiskSpace = isRunningOutOfDiskSpace();
 
+      // Determine whether the node is running out of disk space.
```
johnou: Running out of memory?

johnou:
@kuujo iirc you were happy with this PR after the suggestions were fixed, if not let me know and I will revert.

PR #675, opened 2018-07-05
Fix javadoc generation.

johnou:


kuujo:
This still doesn't generate the aggregated Javadocs... only the docs for each module.

johnou:
@kuujo what if you use the javadoc:aggregate mojo?

kuujo:
Should be fine now

protocols/raft/src/main/java/io/atomix/protocols/raft/ReadConsistency.java
```diff
@@ -15,10 +15,12 @@
  */
 package io.atomix.protocols.raft;
 
+import io.atomix.protocols.raft.protocol.QueryRequest;
+
 /**
- * Constants for specifying Raft {@link RaftQuery} consistency levels.
+ * Constants for specifying Raft {@link QueryRequest} consistency levels.
```
kuujo: Should probably use `OperationType#QUERY`. `QueryRequest` is an internal API

PR #677, opened 2018-07-06
Simple distributed maps

kuujo:
This PR improves all map-related interfaces to make them consistent with core Java/Guava interfaces. It adds `DistributedMap`, `DistributedSortedMap`, `DistributedNavigableMap`, `DistributedTreeMap` and `DistributedMultimap` interfaces/primitives which differ from their `Atomic*` cousins in that they actually implement `Map`, `SortedMap`, `NavigableMap`, `Multimap`, etc. The implementations simply wrap `Atomic*` primitives.

A lot of the changes also ensure views are supported for all the map primitives as well. For example, `DistributedSortedMap#keySet()` returns a `DistributedSortedSet` and `AsyncDistributedSortedMap#keySet()` returns an `AsyncDistributedSortedSet` which is a view of the map's keys. This makes the map/collection primitives much richer.

Many of the actual views are not implemented in this PR, but iterators and streams are, so `map.keySet().stream()` actually lazily streams map keys to the client. The asynchronous primitive APIs also have an improved `AsyncIterator` API.


core/src/main/java/io/atomix/core/collection/impl/PartitionedProxyIterator.java
```diff
@@ -0,0 +1,78 @@
...
+    this.client = client;
+    this.partitions = client.getPartitionIds().iterator();
+    this.openFunction = openFunction;
+    this.nextFunction = nextFunction;
+    this.closeFunction = closeFunction;
+  }
+
+  @Override
+  public CompletableFuture<Boolean> hasNext() {
+    if (iterator == null && partitions.hasNext()) {
```
johnou: iterator is volatile but shouldn't we be using double checked pattern here? or better yet holder-class pattern?

core/src/main/java/io/atomix/core/collection/impl/PartitionedProxyIterator.java
```diff
@@ -0,0 +1,78 @@
...
+            return hasNext();
+          }
+          return CompletableFuture.completedFuture(true);
+        });
+  }
+
+  @Override
+  public CompletableFuture<T> next() {
+    if (iterator == null && partitions.hasNext()) {
+      iterator = new ProxyIterator<>(client, partitions.next(), openFunction, nextFunction, closeFunction);
```
johnou: same here.

core/src/main/java/io/atomix/core/collection/impl/PartitionedProxyIterator.java
kuujo: Actually, the initial iterator should probably be a completed iterator to force it to be populated in the `thenCompose` callback

core/src/main/java/io/atomix/core/iterator/impl/ProxyIterator.java
```diff
@@ -0,0 +1,117 @@
...
+      ProxyClient<S> client,
+      PartitionId partitionId,
+      OpenFunction<S> openFunction,
+      NextFunction<S, T> nextFunction,
+      CloseFunction<S> closeFunction) {
+    this.client = client;
+    this.partitionId = partitionId;
+    this.nextFunction = nextFunction;
+    this.closeFunction = closeFunction;
+    this.openFuture = client.applyOn(partitionId, openFunction::open);
```
johnou: is this an ordered future?

core/src/main/java/io/atomix/core/collection/impl/DefaultDistributedCollectionService.java
```diff
@@ -223,22 +222,54 @@ public void close(long iteratorId) {
   @Override
   public void onExpire(Session session) {
     listeners.remove(session.sessionId());
-    iterators.entrySet().removeIf(entry -> entry.getValue().sessionId == session.sessionId().id());
+    iterators.entrySet().removeIf(entry -> entry.getValue().sessionId() == session.sessionId().id());
```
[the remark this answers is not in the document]
kuujo: yeah... `entry.getValue()` is a `long`

core/src/main/java/io/atomix/core/iterator/impl/ProxyIterator.java
kuujo: that is probably a good point

core/src/main/java/io/atomix/core/map/AsyncAtomicNavigableMap.java
```diff
@@ -0,0 +1,335 @@
...
+  CompletableFuture<Map.Entry<K, Versioned<V>>> pollFirstEntry();
+
+  /**
+   * Removes and returns a key-value mapping associated with
+   * the greatest key in this map, or {@code null} if the map is empty.
+   *
+   * @return the removed last entry of this map,
+   *         or {@code null} if this map is empty
+   */
+  CompletableFuture<Map.Entry<K, Versioned<V>>> pollLastEntry();
```
johnou: wasn't there problems with poll*?

core/src/main/java/io/atomix/core/map/AsyncDistributedMap.java
```diff
@@ -0,0 +1,447 @@
...
+   */
+  CompletableFuture<Void> addListener(MapEventListener<K, V> listener, Executor executor);
+
+  /**
+   * Unregisters the specified listener such that it will no longer receive map change notifications.
+   *
+   * @param listener listener to unregister
+   *
+   * @return future that will be completed when the operation finishes
+   */
```
johnou: impressive javadocs in this class.

core/src/main/java/io/atomix/core/map/impl/AbstractAtomicMapService.java
```diff
@@ -0,0 +1,1058 @@
...
+import java.util.stream.Collectors;
+
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * State Machine for {@link AtomicMapProxy} resource.
+ */
+public abstract class AbstractAtomicMapService<K> extends AbstractPrimitiveService<AtomicMapClient> implements AtomicMapService<K> {
+
+  private static final int MAX_ITERATOR_BATCH_SIZE = 1024 * 32;
```
johnou: batch size too large?

core/src/main/java/io/atomix/core/map/AsyncAtomicNavigableMap.java
kuujo: Yeah, it has to do with partitioned maps. Since `poll*` removes the first/last entry, running it against all the partitions will remove an entry per partition. Since we can only determine which was the lowest/highest entry _after_ it's already been removed, there's not really a way to do this atomically. But since ordered maps now have iterators, they have to be in a single partition now anyways. It's too difficult to maintain order across partitions for iterators.

core/src/main/java/io/atomix/core/map/impl/AbstractAtomicMapService.java
kuujo: Ahh yeah, I think this needs to be changed. It was in bytes before, but it's not in entries IIRC.

kuujo: Nvm no it isn't.

PR #678, opened 2018-07-06
Correct cache config definition.

johnou:


core/src/test/resources/test.conf
```diff
@@ -67,7 +67,9 @@ primitives.foo {
 
 primitives.bar {
   type: set
-  cache.enabled: true
```
johnou: guess this syntax might work too.

kuujo: it works in Typesafe Config

kuujo: Should change these back. I'd like to keep them as examples for users. This test passes fine.

johnou: changed this one back, pretty sure the other one is broken and it's nice to show different syntax imo.

PR #695, opened 2018-07-12
Cache address instances by host/port

kuujo:


utils/src/main/java/io/atomix/utils/net/Address.java
```diff
@@ -172,14 +201,20 @@ public Type type() {
...
   @Override
   public String toString() {
-    switch (type) {
-      case IPV4:
-        return String.format("%s:%d", address().getHostName(), port());
-      case IPV6:
-        return String.format("[%s]:%d", address().getHostName(), port());
-      default:
-        throw new AssertionError();
+    String toString = toStringCache.get(this);
```
johnou: Called often?

kuujo: When logging is enabled

PR #700, opened 2018-07-14
Support key type/value type/element type registrations for primitives

kuujo:
This PR adds `with*Type` methods to collection and map builders to facilitate registration of serializable types.

### DIFF
--- a/core/src/main/java/io/atomix/core/collection/DistributedCollectionBuilder.java
+++ b/core/src/main/java/io/atomix/core/collection/DistributedCollectionBuilder.java
@@ -15,9 +15,14 @@
  */
 package io.atomix.core.collection;
 
+import com.google.common.collect.Lists;
 import io.atomix.primitive.PrimitiveBuilder;
 import io.atomix.primitive.PrimitiveManagementService;
 import io.atomix.primitive.PrimitiveType;
+import io.atomix.utils.serializer.Namespace;
+import io.atomix.utils.serializer.NamespaceConfig;
+import io.atomix.utils.serializer.Namespaces;
+import io.atomix.utils.serializer.Serializer;
 
 /**
  * Distributed collection builder.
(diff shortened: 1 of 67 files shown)


core/src/main/java/io/atomix/core/collection/DistributedCollectionBuilder.java
```diff
@@ -30,4 +35,117 @@
...
   }
+
+  /**
+   * Sets the element type.
+   *
+   * @param elementType the element type
+   * @return the collection builder
+   */
+  @SuppressWarnings("unchecked")
+  public B withElementType(Class<?> elementType) {
```
johnou: Serializable?

core/src/main/java/io/atomix/core/collection/DistributedCollectionBuilder.java
```diff
@@ -30,4 +35,117 @@
...
+          .nextId(Namespaces.BEGIN_USER_CUSTOM_ID + 100);
+
+      namespaceBuilder.setRegistrationRequired(config.isRegistrationRequired());
+      namespaceBuilder.setCompatible(config.isCompatibleSerialization());
+
+      if (config.getElementType() != null) {
+        namespaceBuilder.registerSubTypes(config.getElementType());
+      }
+      if (!config.getExtraTypes().isEmpty()) {
+        namespaceBuilder.register(config.getExtraTypes().toArray(new Class<?>[config.getExtraTypes().size()]));
```
johnou: Better to use zero here instead of exact size.

johnou: Using toArray(new T[0]) is usually the faster

See https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_conclusion

PR #704, opened 2018-07-16
Ensure event listeners are called after cache update

kuujo:
This PR fixes a bug in cached primitives wherein event listeners can be called before the cache has been updated. It intercepts added listeners at the cache layer and calls them only after events update the cache. This is a bug that was discovered in ONOS.


kuujo:
BTW it's probably easiest to look at each of the commits individually.

Issue #703, opened 2018-07-16
getLeaderElector in 3.0.0.rc3 throws ClassCastException


tomasol:
**Expected behavior**

Calling ` atomix.getLeaderElector("test");`, I expect to get an instance or a reasonable exception in case of wrong configuration.

**Actual behavior**
After RAFT is initialized, I get
```
Exception in thread "main" java.lang.ClassCastException: io.atomix.core.election.impl.TranscodingAsyncLeaderElector cannot be cast to io.atomix.primitive.SyncPrimitive
	at io.atomix.core.PrimitivesService.getPrimitive(PrimitivesService.java:763)
	at io.atomix.core.impl.CorePrimitivesService.getLeaderElector(CorePrimitivesService.java:277)
	at io.atomix.core.Atomix.getLeaderElector(Atomix.java:467)
	at foo.AtomixTest.main(AtomixTest.java:48)
```

**Steps to reproduce**
Since the documentation is lagging, I butchered following code to test RAFT cluster:

```java

    public static void main(String[] args) throws InterruptedException {
        int idx = Integer.parseInt(args[0]);
        int size = Integer.parseInt(args[1]);
        if (idx < 0 || idx >= size) {
            throw new IllegalStateException();
        }

        AtomixBuilder builder = Atomix.builder();
        Function<Integer, Integer> idxToPort = i -> 5000 + i;
        IntFunction<String> idxToMemberName = i -> "member-" + i;
        builder.withAddress(idxToPort.apply(idx));
        builder.withMemberId(idxToMemberName.apply(idx));

        BootstrapDiscoveryBuilder bootstrapDiscoveryBuilder = BootstrapDiscoveryProvider.builder();

        IntFunction<Node> idxToNode = i -> Node.builder().withId(idxToMemberName.apply(i)).withAddress(idxToPort.apply(i)).build();
        bootstrapDiscoveryBuilder.withNodes(IntStream.range(0, size).mapToObj(idxToNode).toArray(Node[]::new));
        builder.withMembershipProvider(bootstrapDiscoveryBuilder.build());

        ConsensusProfileBuilder consensusProfileBuilder = ConsensusProfile.builder();
        consensusProfileBuilder.withNumPartitions(1);

        Set<String> members = IntStream.range(0, size).mapToObj(idxToMemberName).collect(Collectors.toSet());

        builder.withProfiles(consensusProfileBuilder.withMembers(members).build());
        Atomix atomix = builder.build();
        atomix.start().join();

[...]

PR #710, opened 2018-07-18
Add primitive getter and builder tests

kuujo:
This PR adds various tests for primitive getters and builders. It tests all specific primitive getters, abstract pre-configured primitive getters, and primitive builder methods to ensure all methods return valid and correct primitive instances.


kuujo:
#703

Issue #712, opened 2018-07-18
Atomix start: 5 seconds spent on getHostName of a multicast IP address


imperatorx:
**Expected behavior**

Atomix starts up in a timely manner

**Actual behavior**

Atomix waits for ~5 sec here. Should the call in AtomixCluster be 

`config.getMulticastConfig().getGroup().getHostAddress()`

instead of 

`config.getMulticastConfig().getGroup().getHostName()`?

```
	at java.net.Inet6AddressImpl.getHostByAddr(java.base@10/Native Method)
	at java.net.InetAddress$PlatformNameService.getHostByAddr(java.base@10/InetAddress.java:931)
	at java.net.InetAddress.getHostFromNameService(java.base@10/InetAddress.java:655)
	at java.net.InetAddress.getHostName(java.base@10/InetAddress.java:598)
	at java.net.InetAddress.getHostName(java.base@10/InetAddress.java:570)
	at io.atomix.cluster.AtomixCluster.buildBroadcastService(AtomixCluster.java:327)
	at io.atomix.cluster.AtomixCluster.<init>(AtomixCluster.java:184)
	at io.atomix.core.Atomix.<init>(Atomix.java:269)
	at io.atomix.core.AtomixBuilder.build(AtomixBuilder.java:264)
```

**Steps to reproduce**

Start with code below

**Minimal yet complete reproducer code (or URL to code)**

```java
     Member node = Member.builder()
                .withId(MemberId.from("1"))
                .withAddress(currentHost, currentPort)
                .build();

        Atomix atomix = Atomix.builder()
                .withAddress(currentHost, currentPort)
                .withMembershipProvider(new BootstrapDiscoveryProvider(node))
                .withMulticastEnabled(false)
                .withProfiles(Profile.dataGrid(5))
                .withClusterId("Cluster 1")
                .build();

        atomix.start().join();
```

**Environment**

 - Atomix: 3.0.0-rc3
 - OS: Windows 7
 - JVM 10

<sup>:warning:️ Please verify that your issue still occurs on the latest version of Atomix before reporting. 
The documentation is currently work-in-progress and is not yet complete.</sup>

-----------------------------------------------------------------------------------------
*Have you searched the CLOSED issues already? How about checking in with the Atomix [google group](https://groups.google.com/forum/#!forum/atomixio)?*

johnou:
@imperatorx can you try passing in the following JVM opt? `-Djava.net.preferIPv4Stack=true`

[...]

PR #713, opened 2018-07-18
Do not resolve multicast IP address to host to speed up startup

imperatorx:
Im my case it tried to resolve 230.0.0.1 to a hostname.


imperatorx:
#712

PR #715, opened 2018-07-18
Fix various bugs in REST API tests

kuujo:


kuujo:
Seems like the tests passed but the shaded jar failed for some reason

Issue #722, opened 2018-07-21
memory mapped segment  compacting  throw exception when closeing MappedBytes & deleting file


aruanruan:
java.lang.RuntimeException: java.nio.file.FileSystemException: target\perf-logs\1\atomix-1.log: The process cannot access the file because it is being used by another process
        at io.atomix.storage.buffer.MappedBytes.delete(MappedBytes.java:xxx) ~[classes/:na]
	at io.atomix.storage.buffer.MappedBuffer.delete(MappedBuffer.java:xxx) ~[classes/:na]
	at io.atomix.storage.journal.JournalSegmentWriter.delete(JournalSegmentWriter.java:263) ~[classes/:na]
	at io.atomix.storage.journal.JournalSegment.delete(JournalSegment.java:190) ~[classes/:na]
	at io.atomix.storage.journal.SegmentedJournal.compact(SegmentedJournal.java:643) ~[classes/:na]
	at io.atomix.protocols.raft.storage.log.RaftLog.compact(RaftLog.java:128) ~[classes/:na]
	at io.atomix.protocols.raft.impl.RaftServiceManager.lambda$7(RaftServiceManager.java:291) ~[classes/:na]
	at java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source) ~[na:1.8.0_144]

THIS is a very old BUG: 

https://bugs.java.com/view_bug.do?bug_id=4829325

 
I tried some solutions : 
https://qiita.com/ajis_ka/items/e89cd2472e961bac69bc
https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java#L339

but it does not WORK yet

**Environment**

 - Atomix: 3.0.0
 - OS:  windows 10
 - JVM 1.8.0-131

<sup>:warning:️ Please verify that your issue still occurs on the latest version of Atomix before reporting. 
The documentation is currently work-in-progress and is not yet complete.</sup>

-----------------------------------------------------------------------------------------
*Have you searched the CLOSED issues already? How about checking in with the Atomix [google group](https://groups.google.com/forum/#!forum/atomixio)?*

johnou:
We free the direct buffer here https://github.com/atomix/atomix/blob/b48608543581dbc1ba56f568efa199c377a993a3/utils/src/main/java/io/atomix/utils/memory/MappedMemory.java#L80

johnou:
[...]

PR #728, opened 2018-07-22
Use Lucene cleaner and free mappedbytes buffer on close.

johnou:
Fixes https://github.com/atomix/atomix/issues/722


storage/src/main/java/io/atomix/storage/buffer/MappedBytes.java
```diff
@@ -105,6 +106,7 @@ public Bytes flush() {
   @Override
   public void close() {
     try {
+      BufferCleaner.CLEANER.freeBuffer(buffer);
```
johnou: Might make sense to try catch free buffer and ignore fail then close the RAF.

utils/src/main/java/io/atomix/utils/memory/BufferCleaner.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.Objects;
+
+import static java.lang.invoke.MethodHandles.constant;
+import static java.lang.invoke.MethodHandles.dropArguments;
+import static java.lang.invoke.MethodHandles.filterReturnValue;
+import static java.lang.invoke.MethodHandles.guardWithTest;
+import static java.lang.invoke.MethodHandles.lookup;
+import static java.lang.invoke.MethodType.methodType;
+
+public class BufferCleaner {
```
kuujo: Public undocumented class

johnou: Do we have a way to make it internal but keep it public since it's needed in multiple packages?

johnou:
@kuujo PTAL Atomix now compiles on Java 11.

storage/src/main/java/io/atomix/storage/buffer/DirectBytes.java
```diff
@@ -33,9 +31,9 @@
    *                                  an array on the Java heap - {@code Integer.MAX_VALUE - 5}
    */
   public static DirectBytes allocate(int size) {
-    if (size > HeapMemory.MAX_SIZE)
-      throw new IllegalArgumentException("size cannot for DirectBytes cannot be greater than " + HeapMemory.MAX_SIZE);
-    return new DirectBytes(ByteBuffer.allocate((int) size));
+    if (size > MAX_SIZE)
+      throw new IllegalArgumentException("size cannot for DirectBytes cannot be greater than " + MAX_SIZE);
+    return new DirectBytes(ByteBuffer.allocateDirect((int) size));
```
johnou: DirectBytes allocate was allocating heap bytebuffer..

storage/src/main/java/io/atomix/storage/buffer/DirectBytes.java
```diff
@@ -49,7 +47,7 @@ protected ByteBuffer newByteBuffer(int size) {
 
   @Override
   public boolean isDirect() {
-    return true;
+    return buffer.isDirect();
```
johnou: so we can verify in unit test :ok_hand:

johnou:
@kuujo PTAL

PR #742, opened 2018-07-28
Fix failure in mapped segment journal compaction without deleting log files

aruanruan:
- Fix failure in mapped segment journal compaction without deleting log files
- Fix  failure in  raft performance test


johnou:
Closing in favour of https://github.com/atomix/atomix/pull/728

PR #743, opened 2018-07-29
Throw exception on duplicate storage directory in Raft servers

kuujo:
This PR is an attempt to disambiguate Raft partition configurations by throwing exceptions when servers on different nodes are configured to use the same data directory. This is accomplished by storing a `.partition-n.lock` file which contains the local server ID. When a Raft server context is created, the constructor validates that its storage directory does not conflict with another node's storage directory.

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftContext.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftContext.java
@@ -42,6 +42,7 @@ import io.atomix.protocols.raft.storage.log.RaftLogWriter;
 import io.atomix.protocols.raft.storage.snapshot.SnapshotStore;
 import io.atomix.protocols.raft.storage.system.MetaStore;
 import io.atomix.protocols.raft.utils.LoadMonitor;
+import io.atomix.storage.StorageException;
 import io.atomix.utils.concurrent.SingleThreadContext;
 import io.atomix.utils.concurrent.ThreadContext;
 import io.atomix.utils.concurrent.ThreadContextFactory;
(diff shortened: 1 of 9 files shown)


protocols/raft/src/main/java/io/atomix/protocols/raft/storage/RaftStorage.java
```diff
@@ -227,6 +229,37 @@ public StorageStatistics statistics() {
     return statistics;
   }
 
+  /**
+   * Attempts to acquire a lock on the storage directory.
+   *
+   * @param id the ID with which to lock the directory
+   * @return indicates whether the lock was successfully acquired
+   */
+  public boolean lock(String id) {
```
johnou: what about something like this? https://stackoverflow.com/a/9020391/7534815

kuujo: Not necessary in this case. It's overkill. We're not trying to control concurrent access to a shared file. All we need is to ensure only a single file can operate on a log directory, and this accomplishes that. The `File.createNewFile` call atomically creates a file. If the file does not exist, the first node to create the file will effectively acquire a persistent, forever lock on the data directory. If it does exist, the node just must read the file to verify that it's the one that created it in the past. There's no need to use channel locks because this lock is never actually released for the purposes of allowing another node to acquire it.

kuujo:
We have a problem with the test logs using the 4MB of space Travis allows

protocols/raft/src/main/java/io/atomix/protocols/raft/storage/RaftStorage.java
johnou: We could release it immediately, point was to address the race you mentioned.

kuujo:
This seems to be causing problems.

kuujo:
Many of the core tests were not using correctly isolated storage directories. Curious to see how they change after fixing that problem.

kuujo:
Need to make sure this exception is propagated or logged

protocols/raft/src/main/java/io/atomix/protocols/raft/impl/RaftContext.java
```diff
@@ -875,6 +881,9 @@ public void delete() {
 
     // Delete the metadata store.
     storage.deleteMetaStore();
+
+    // Unlock the store.
+    storage.unlock();
```
johnou: finally block?

PR #745, opened 2018-07-30
Segment size in RaftPartitionGroupBuilder

alex-richter:
Allowed the segment size of raft partitions to be changed through the RaftPartitionGroupBuilder.


johnou:
@alex-richter thanks!

kuujo:
Thanks 👏

PR #747, opened 2018-07-30
Ensure primitives are correctly filtered by type when reading primitive info

kuujo:


core/src/main/java/io/atomix/core/impl/CorePrimitiveRegistry.java
```diff
@@ -87,7 +87,7 @@ public CorePrimitiveRegistry(PartitionService partitionService, PrimitiveTypeReg
   public Collection<PrimitiveInfo> getPrimitives(PrimitiveType primitiveType) {
     return getPrimitives()
         .stream()
-        .filter(primitive -> primitive.type().equals(primitiveType))
+        .filter(primitive -> primitive.type().name().equals(primitiveType.name()))
```
johnou: What was equals by type returning?

PR #749, opened 2018-08-01
Change static factory methods for DocumentPath

kuujo:
Construct a root path when an array/list of nodes is provided. Based on discussion with @aruanruan

### DIFF
--- a/core/src/main/java/io/atomix/core/tree/DocumentPath.java
+++ b/core/src/main/java/io/atomix/core/tree/DocumentPath.java
@@ -18,7 +18,6 @@ package io.atomix.core.tree;
 
 import com.google.common.collect.Comparators;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Lists;
 import org.apache.commons.lang3.StringUtils;
 
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/atomix/core/tree/DocumentPath.java
```diff
@@ -36,20 +35,67 @@
...
+
+  /**
+   * Creates a new root {@code DocumentPath} from a list of path elements.
+   *
+   * @param elements path elements
+   * @return {@code DocumentPath} instance
+   */
+  public static DocumentPath from(List<String> elements) {
+    List<String> copy = new ArrayList<>();
+    copy.add("");
```
johnou: what's the empty string for?

johnou:
```java
  /**
   * Root document path.
   */
  private static final String ROOT_PATH = "/";

  /**
   * Root document tree path.
   */
  public static final DocumentPath ROOT = new DocumentPath(ImmutableList.of(""));
```
do these still match?

johnou: just read over the converation on gitter, might be nicer if we introduce a constant here, ROOT_ELEMENT, self documenting code ftw.

kuujo: Good plan

PR #755, opened 2018-08-02
Filter bootstrap heartbeats by local node identifier

kuujo:
Rather than by `Address`


kuujo:
Hmm I'm not sure this is behaving well.

kuujo:
Going to close this and investigate it some more in ONOS.

PR #756, opened 2018-08-02
Block startup on partition group discovery to avoid race conditions

kuujo:


kuujo:
An exception is only caught and returned if it’s unexpected, i.e. a bug. Then startup is failed immediately as it should be.

PR #757, opened 2018-08-02
Use agent distribution in Docker image

kuujo:
Updates the Dockerfile to use the Atomix agent distribution. The distro needs to be updated separately to support profiling.


Dockerfile
```diff
@@ -6,30 +6,13 @@ MAINTAINER  Jordan Halterman <jordan.halterman@gmail.com>
 
 RUN apt-get update && apt-get install -y iptables git stress
```
johnou: Openjdk?

kuujo: This image extends the `java:8` image

PR #762, opened 2018-08-03
Create separate directory in distribution for user extensions

kuujo:
This PR moves Atomix jars into the `/system` folder and preserves the `/lib` folder for user extensions (custom primitives, protocols, etc).


kuujo:
Maybe it should be `/lib/system` and `/lib/ext`

PR #764, opened 2018-08-03
Provisional JDK10+ support.

johnou:
Unit tests should now pass on Java 10+


johnou:
@kuujo I had meant that I was able to reproduce the Multiset bug in Java 8 like in Java 10, PTAL.

PR #768, opened 2018-08-06
Use fixed window rolling policy to ensure logs roll over on size limit

kuujo:
Update the distro's default logback configuration to ensure log files are rolled over correctly on size limits.

### DIFF
--- a/dist/src/main/resources/conf/logback.xml
+++ b/dist/src/main/resources/conf/logback.xml
@@ -17,9 +17,10 @@
     <appender name="FILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <file>${atomix.log}/atomix.log</file>
 
-        <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
-            <fileNamePattern>atomix.%d{yyyy-MM-dd}.log</fileNamePattern>
-            <maxHistory>10</maxHistory>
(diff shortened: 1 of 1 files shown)


dist/src/main/resources/conf/logback.xml
```diff
@@ -17,9 +17,10 @@
     <appender name="FILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <file>${atomix.log}/atomix.log</file>
 
-        <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
-            <fileNamePattern>atomix.%d{yyyy-MM-dd}.log</fileNamePattern>
-            <maxHistory>10</maxHistory>
+        <rollingPolicy class="ch.qos.logback.core.rolling.FixedWindowRollingPolicy">
+            <fileNamePattern>${atomix.log}/atomix.log.%i</fileNamePattern>
```
johnou: Are you sure the %1 goes at the end? https://logging.apache.org/log4j/extras/apidocs/org/apache/log4j/rolling/FixedWindowRollingPolicy.html

kuujo: What do you mean? Am I sure it works at the end? Yes.

johnou: ah okay in the documentation they say, somewhere, just that their example is "foo.%i.log".

PR #770, opened 2018-08-07
WIP: Use Netty SslContextBuilder for configuring TLS.

johnou:
Still need to test with actual certificates.


cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyMessagingService.java
```diff
@@ -698,21 +711,18 @@ protected void initChannel(SocketChannel channel) throws Exception {
...
-
-      SSLEngine clientSslEngine = clientContext.createSSLEngine();
-
-      clientSslEngine.setUseClientMode(true);
-      clientSslEngine.setEnabledProtocols(clientSslEngine.getSupportedProtocols());
-      clientSslEngine.setEnabledCipherSuites(clientSslEngine.getSupportedCipherSuites());
-      clientSslEngine.setEnableSessionCreation(true);
-
-      channel.pipeline().addLast("ssl", new io.netty.handler.ssl.SslHandler(clientSslEngine))
+      channel.pipeline().addLast("ssl", sslContext.newHandler(channel.alloc(), address.host(), address.port()))
```
johnou: specifying the host and port is important for SNI to work.

PR #772, opened 2018-08-07
Flush badge cache

johnou:


kuujo:
On an unrelated note, can we change the coveralls failure threshold?

Issue #780, opened 2018-08-09
Limit Raft log entry size


kuujo:
Limit the maximum size of Raft log entries. This is necessary to avoid allocating too much memory when reading corrupted entries after a node crash.

PR #793, opened 2018-08-09
Implement Raft log entry size limit

kuujo:
#780 

Imposes a maximum Raft log entry size. This is used to avoid allocating too much memory when computing checksums for Raft log entries. This PR also refactors the Raft partition group configuration to create `storage` and `compaction` sub-configurations.

### DIFF
--- a/storage/src/main/java/io/atomix/storage/journal/JournalSegmentWriter.java
+++ b/storage/src/main/java/io/atomix/storage/journal/JournalSegmentWriter.java
@@ -15,6 +15,7 @@
  */
 package io.atomix.storage.journal;
 
+import io.atomix.storage.StorageException;
 import io.atomix.utils.serializer.Serializer;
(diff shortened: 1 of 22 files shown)


johnou:
nodes were not able to reach consensus and continued logging?

storage/src/main/java/io/atomix/storage/journal/JournalSegmentWriter.java
```diff
@@ -191,6 +199,11 @@ public void append(Indexed<E> entry) {
     final byte[] bytes = serializer.encode(entry);
     final int length = bytes.length;
 
+    // If the entry length exceeds the maximum entry size then throw an exception.
+    if (length > maxEntrySize) {
+      throw new StorageException.TooLarge("Entry size cannot exceed " + maxEntrySize + " bytes");
```
johnou: might be useful to include the actual length too.

Issue #790, opened 2018-08-09
Mark experimental interfaces with @Beta annotation


kuujo:
Atomix 3.0.0 will be released with some experimental interfaces. In particular, eventually consistent primitive protocols are likely to be re-architected post-3.0.0. Those interfaces need to be marked with the `@Beta` annotation.

PR #796, opened 2018-08-10
Add @Beta annotation to eventually consistent protocols

kuujo:


kuujo:
Resolves #790

johnou:
@kuujo could just add our own Beta annotation to utils instead of adding more guava?

Issue #789, opened 2018-08-09
Add version compatibility checks


kuujo:
Introduce version numbers and node versions to cluster management to support backwards compatibility checks when establishing connections between nodes. This is necessary to ensure that e.g. a client does not attempt to create a primitive that doesn’t yet exist in an old version of the software.

PR #797, opened 2018-08-10
Add Atomix version utility

kuujo:


kuujo:
This is a precursor to #789

PR #798, opened 2018-08-10
Attach simple numeric version number to Netty messaging protocol

### LINKED ISSUE #787: Version Netty connections
Introduce a version number to intra-cluster Netty connections for protocol compatibility. The connection version can be established on the initial setup when the sender’s address is established.

kuujo:
Fixes #787 

This PR attaches a simple numeric version number to the initialization of connections in the Netty messaging service implementation. This can be used to ensure modifications can be made to the encoders/decoders without breaking backwards compatibility in the future. A change would simply require maintaining versioned encoders/decoders and encoding/decoding messages based on the connection versions.


cluster/src/main/java/io/atomix/cluster/messaging/impl/MessageEncoder.java
```diff
@@ -60,6 +61,8 @@ protected void encode(
   private void encodeMessage(InternalMessage message, ByteBuf out) {
     // If the address hasn't been written to the channel, write it.
     if (!addressWritten) {
+      out.writeByte(VERSION);
```
johnou: short?

kuujo: 128 protocol changes?

PR #799, opened 2018-08-10
Classgraph fixes

johnou:
Update of ClassGraph and closing the ScanResult will result in a lower memory footprint.


kuujo:
K no more dependency changes after this. Already wary of changing this dependency again.

Issue #784, opened 2018-08-09
Support deleting primitives


kuujo:
Add a `delete()` API to all primitives and implement semantics for operating on deleted primitive instances. When a primitive is deleted, the state of the primitive should be removed from disk and memory and all open instances of the primitive should receive exceptions when operating on the primitive.

PR #801, opened 2018-08-11
Support deleting primitive services managed via the Raft consensus protocol

### LINKED ISSUE #782: Support deletes for Raft primitives
Support deleting primitive services in the Raft implementation. This will require introducing new error types for deleted Raft primitives.

kuujo:
This PR is an initial attempt to resolve #782 by implementing deletes in the Raft protocol.

The most significant challenge with implementing deletes is avoiding the recreation of primitive services following a delete. Primitive clients generally create primitive services using a create-if-not-exists write. So, the risk here is that after one session deletes the primitive service, another process recognizes its session has been closed and simply recreates the session, thus recreating the primitive service. Instead, when a primitive is deleted by one process, all other processes should see that the primitive has been deleted.

To ensure this is the case, the sessions that belonged to a primitive service are retained after the service itself has been deleted. Those sessions are retained in memory for as long as their session timeout. When an operation is attempted on those sessions, an `UnknownService` exception is returned, thus notifying the client that the service has been deleted rather than that its session was expired or closed.

To facilitate session recovery, an additional `PrimitiveState` is added to differentiate between `EXPIRED` sessions and `CLOSED` sessions, the latter being the state that occurs after a client's session is closed by a delete. This allows the `RecoveringSessionClient` to avoid recreating a session after a primitive has been deleted.

### DIFF
--- a/core/src/main/java/io/atomix/core/barrier/impl/BlockingDistributedCyclicBarrier.java
+++ b/core/src/main/java/io/atomix/core/barrier/impl/BlockingDistributedCyclicBarrier.java
@@ -15,6 +15,7 @@
  */
 package io.atomix.core.barrier.impl;
 
+import com.google.common.base.Throwables;
 import io.atomix.core.barrier.AsyncDistributedCyclicBarrier;
 import io.atomix.core.barrier.DistributedCyclicBarrier;
 import io.atomix.primitive.PrimitiveException;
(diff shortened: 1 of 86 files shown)


kuujo:
Also resolves #784, although we need to add more tests before that issue is closed

johnou:
So if I understand correctly only existing sessions will receive unknown service when it's deleted so if it is "created" after deletion that will work correctly.

kuujo:
Yes. Need to add more tests to recreate a primitive, plus add delete tests to each individual primitive just to be sure.

kuujo:
Added more changes to ensure the primitive registry is updated and tests for all core primitives

PR #803, opened 2018-08-11
Simplify the core Docker image to reduce size

### LINKED ISSUE #802: Simplify Docker image
The Docker image currently contains some packages that are not necessary to run Atomix. The core image should be reduced to only the bare minimum necessary features. After all, that's how Docker is intended to be used.

kuujo:
This came from the Sona project:
https://github.com/sonaproject/atomix-docker
https://hub.docker.com/r/opensona/atomix-docker/tags/

Fixes #802 

Need to test the image with https://github.com/atomix/atomix-test before merging

### DIFF
--- a/dist/Dockerfile
+++ b/dist/Dockerfile
@@ -1,18 +1,15 @@
 # Dockerfile
 
-FROM  java:8
-
-MAINTAINER  Jordan Halterman <jordan.halterman@gmail.com>
-
-RUN apt-get update && apt-get install -y iptables git stress
-
-RUN mkdir /atomix
+FROM anapsix/alpine-java:8_server-jre
 
 ARG VERSION
-COPY target/atomix-agent-${VERSION}.tar.gz /atomix/atomix-agent-${VERSION}.tar.gz
-RUN tar -xvf /atomix/atomix-agent-${VERSION}.tar.gz -C /atomix && rm /atomix/atomix-agent-${VERSION}.tar.gz
+RUN mkdir -p /root/atomix
+COPY target/atomix-agent-${VERSION}.tar.gz /root/atomix/atomix-agent-${VERSION}.tar.gz
+RUN tar -xvf /root/atomix/atomix-agent-${VERSION}.tar.gz -C /root/atomix && rm /root/atomix/atomix-agent-${VERSION}.tar.gz
+
+WORKDIR /root/atomix
 
 EXPOSE 5678
 EXPOSE 5679
 
-ENTRYPOINT ["/atomix/bin/atomix-agent"]
+ENTRYPOINT ["./bin/atomix-agent"]


kuujo:
The Dockerfile is also missing the license/copyright header

dist/Dockerfile
```diff
@@ -1,18 +1,18 @@
...
-RUN mkdir /atomix
+MAINTAINER  Jordan Halterman <jordan@opennetworking.org>
 
 ARG VERSION
-COPY target/atomix-agent-${VERSION}.tar.gz /atomix/atomix-agent-${VERSION}.tar.gz
-RUN tar -xvf /atomix/atomix-agent-${VERSION}.tar.gz -C /atomix && rm /atomix/atomix-agent-${VERSION}.tar.gz
+RUN mkdir -p /root/atomix
+COPY target/atomix-agent-${VERSION}.tar.gz /root/atomix/atomix-agent-${VERSION}.tar.gz
+RUN tar -xvf /root/atomix/atomix-agent-${VERSION}.tar.gz -C /root/atomix
+RUN rm /root/atomix/atomix-agent-${VERSION}.tar.gz
```
johnou: better to chain both of these run commands with && to reduce the amount of layers.

kuujo: Ahh yes I knew there was some reason to do that

dist/Dockerfile
```diff
@@ -1,18 +1,18 @@
...
 
-FROM  java:8
+FROM anapsix/alpine-java:8_server-jre
 
-MAINTAINER  Jordan Halterman <jordan.halterman@gmail.com>
-
-RUN apt-get update && apt-get install -y iptables git stress
-
-RUN mkdir /atomix
+MAINTAINER  Jordan Halterman <jordan@opennetworking.org>
```
jplock: Specifying the maintainer is deprecated

dist/Dockerfile
```diff
@@ -1,18 +1,18 @@
 # Dockerfile
 
-FROM  java:8
+FROM anapsix/alpine-java:8_server-jre
```
jplock: Why use this image as a base instead of one of the officially supported openjdk alpine base images?

johnou: also keep in mind that if you decide to use ssl in the tcp connection the JDK implementation generates a lot of garbage and there are no alpine artifacts for netty-tcnative.

johnou: @kuujo try `FROM openjdk/8-jdk-slim` compressed size is 91 M.

https://hub.docker.com/r/library/openjdk/tags/

johnou: @kuujo try `FROM openjdk/8-jdk-alpine` it's an official image.

PR #806, opened 2018-08-16
Read/write Raft logs in batches

kuujo:
This PR includes several commits that optimize reading and writing the Raft log:
* Avoid system calls by reading and writing entries as a single array
* Pre-allocate contiguous Raft log segments
* Avoid cache misses and disk reads on Raft log tail reads
* Ensure cluster events are handled on the Raft server thread

The last point is a potential fix for the `Cannot apply index` issue in `RaftServiceManager`

### DIFF
--- a/storage/src/main/java/io/atomix/storage/journal/JournalSegmentReader.java
+++ b/storage/src/main/java/io/atomix/storage/journal/JournalSegmentReader.java
@@ -38,7 +38,7 @@ public class JournalSegmentReader<E> implements JournalReader<E> {
   private final JournalSegmentCache cache;
   private final JournalIndex index;
   private final Serializer serializer;
-  private final HeapBuffer memory = HeapBuffer.allocate();
+  private final Buffer memory;
   private final long firstIndex;
   private Indexed<E> currentEntry;
   private Indexed<E> nextEntry;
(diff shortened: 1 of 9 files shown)


kuujo:
Have to merge #793 before reviewing this

storage/src/main/java/io/atomix/storage/journal/JournalSegmentReader.java
```diff
@@ -136,7 +136,7 @@ private void readNext() {
     Indexed cachedEntry = cache.get(index);
     if (cachedEntry != null) {
       this.nextEntry = cachedEntry;
-      buffer.skip(cachedEntry.size() + Bytes.INTEGER + Bytes.INTEGER);
+      buffer.skip(memory.position() + cachedEntry.size() + Bytes.INTEGER + Bytes.INTEGER);
```
johnou: @kuujo is it feasible to introduce some unit tests to assert the correct behaviour?

kuujo: I will be adding unit tests before this is merged, but some of these cases are pretty complicated

kuujo: Actually, just setting the cache size to `1` reproduces this in tests

kuujo:
Merging this to get it in an RC so we can test it tomorrow

PR #807, opened 2018-08-17
Use official Docker image

johnou:
For security etc.

### DIFF
--- a/dist/Dockerfile
+++ b/dist/Dockerfile
@@ -1,6 +1,6 @@
 # Dockerfile
 
-FROM anapsix/alpine-java:8_server-jre
+FROM openjdk/8-jdk-alpine
 
 ARG VERSION
 RUN mkdir -p /root/atomix


kuujo:
Changed to `openjdk:8-jre-alpine` but it doesn't run at all

MiniDigger:
there seems to be an issue with their build system, the 8-jre-alpine tag doesn't show under tags (https://hub.docker.com/r/library/openjdk/tags/) , but the docker file is the repo (https://github.com/docker-library/openjdk/blob/master/8/jre/alpine/Dockerfile) and the tag is mentioned in their readme (https://hub.docker.com/_/openjdk/).

kuujo:
This is all I get when I run the container:
> standard_init_linux.go:190: exec user process caused "no such file or directory"

kuujo:
The reason for using the `anapsix` image is because it's small but also includes bash for running setup scripts. The official images do not. The `anapsix` image is not exactly obscure either.

PR #814, opened 2018-08-20
Add build number to filtered resource.

johnou:


core/src/main/java/io/atomix/core/Atomix.java
```diff
@@ -345,6 +352,18 @@ public static AtomixBuilder builder(AtomixConfig config, ClassLoader classLoader
...
 
+  private static final String BUILD;
+  private static final Version VERSION;
+
+  static {
+    try {
+      BUILD = Resources.toString(Atomix.class.getClassLoader().getResource(VERSION_RESOURCE), StandardCharsets.UTF_8);
+    } catch (IOException e) {
+      throw new ConfigurationException("Failed to load Atomix version", e);
+    }
```
johnou: Won't this be thrown when running in IDE

johnou: nope, returns null.

core/src/main/java/io/atomix/core/Atomix.java
```diff
@@ -357,8 +358,9 @@ public static AtomixBuilder builder(AtomixConfig config, ClassLoader classLoader
 
   static {
     try {
-      BUILD = Resources.toString(Atomix.class.getClassLoader().getResource(VERSION_RESOURCE), StandardCharsets.UTF_8);
-    } catch (IOException e) {
+      BUILD = Resources.toString(Objects.requireNonNull(Atomix.class.getClassLoader().getResource(VERSION_RESOURCE),
```
kuujo: Should use `checkNotNull` for consistency

PR #815, opened 2018-08-20
Remove unnecessary version checks when loading Raft log segments

kuujo:


johnou:
@kuujo I see two loops, don't we need to apply the same fix to the one above too?

kuujo:
Not exactly the same. However, because of changes to log compaction some of these checks are probably not even needed any more

kuujo:
I think this needs some more work. The logic for loading log segments is based on assumptions from the old compaction algorithm.

PR #817, opened 2018-08-21
Throw BufferOverflowException when not enough bytes remaining in segment

kuujo:
This PR fixes a bug in the Raft log that causes buffer offsets to be misaligned. The problem is, the `Buffer` implementation does not throw `BufferOverflowException`s when a `Buffer` is written with too many bytes. Instead, it just reads the bytes that it can and modifies the buffer position. So, this change checks the number of bytes being written and throws a `BufferOverflowException` to ensure the log is rolled over to a new segment to write a new entry.

### DIFF
--- a/protocols/raft/src/test/java/io/atomix/protocols/raft/storage/log/AbstractLogTest.java
+++ b/protocols/raft/src/test/java/io/atomix/protocols/raft/storage/log/AbstractLogTest.java
@@ -99,150 +99,148 @@ public abstract class AbstractLogTest {
   @Test
   @SuppressWarnings("unchecked")
   public void testLogWriteRead() throws Exception {
-    RaftLog log = createLog();
-    RaftLogWriter writer = log.writer();
-    RaftLogReader reader = log.openReader(1, RaftLogReader.Mode.ALL);
+    try (RaftLog log = createLog()) {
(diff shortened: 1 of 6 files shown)


protocols/raft/src/test/java/io/atomix/protocols/raft/storage/log/AbstractLogTest.java
```diff
@@ -233,15 +209,15 @@ public void testResetTruncateZero() throws Exception {
     RaftLogReader reader = log.openReader(1, RaftLogReader.Mode.ALL);
 
     assertEquals(0, writer.getLastIndex());
-    writer.append(new InitializeEntry(1, System.currentTimeMillis()));
-    writer.append(new InitializeEntry(1, System.currentTimeMillis()));
+    writer.append(ENTRY);
```
johnou: Does it matter that we no longer increment the second parameter?

kuujo: nope... there was not much point to using real entries. Better to have determinism anyways

kuujo:
The file test seems to not be closing files with this change so need to fix that

PR #819, opened 2018-08-21
Share a single Atomix cluster for primitive tests

kuujo:
This PR is an attempt to improve the speed of tests by sharing a single Atomix cluster and client nodes across all primitive tests rather than setting up and tearing down nodes for every test. After this will come some parallelism, but first...

Let's see how it does!


kuujo:
Well that didn't work

kuujo:
Closing this

PR #826, opened 2018-08-27
Detect and discard misaligned log segments on startup

kuujo:


storage/src/main/java/io/atomix/storage/journal/SegmentedJournal.java
```diff
@@ -512,19 +513,21 @@ synchronized void removeSegment(JournalSegment segment) {
...
+    while (iterator.hasNext()) {
+      JournalSegment<E> segment = iterator.next().getValue();
       if (previousSegment != null && previousSegment.lastIndex() != segment.index() - 1) {
-        log.error("Journal appears to be corrupted! {} is not aligned with the prior segment {}", segment.file().file(), previousSegment.file().file());
+        log.warn("Journal is inconsistent. {} is not aligned with prior segment {}", segment.file().file(), previousSegment.file().file());
         corrupted = true;
       }
+      if (corrupted) {
+        segment.close();
+        segment.delete();
```
johnou: Can delete throw a runtime exception?

PR #827, opened 2018-08-30
Ensure listeners are removed from atomic value on session expire/close

kuujo:
This PR fixes a bug in the `AtomicValue` service that can result in NPEs after a client is shutdown and recreated. The service doesn't correctly remove a session's listeners when it's expired or closed.


core/src/main/java/io/atomix/core/value/impl/AbstractAtomicValueService.java
```diff
@@ -106,4 +107,14 @@ public void addListener() {
   public void removeListener() {
     listeners.remove(getCurrentSession().sessionId());
   }
+
+  @Override
+  protected void onExpire(Session session) {
+    listeners.remove(session.sessionId());
```
johnou: I assume if a client times out and then reconnects the listener will be re-registered?

kuujo: yep

PR #830, opened 2018-09-04
Share thread pool for all partitions in a Raft partition group

kuujo:
Large Raft partition groups currently use far too many threads. The `AtomicMap` performance test creates around a thousand threads. This PR ensures Raft partition clients and servers share a thread pool for each partition group.


protocols/raft/src/main/java/io/atomix/protocols/raft/partition/RaftPartitionGroup.java
```diff
@@ -231,6 +249,10 @@ public RaftPartition getPartition(PartitionId partitionId) {
         .map(RaftPartition::close)
         .collect(Collectors.toList());
     return CompletableFuture.allOf(futures.toArray(new CompletableFuture[futures.size()])).thenRun(() -> {
+      try {
+        threadContextFactory.close();
+      } catch (Exception e) {
```
johnou: What might we be ignoring here? Debug?

PR #842, opened 2018-09-21
avoid freezing

zwillim:
add determine statement to avoid freezing when `eventListeners` is empty


johnou:
@zwillim ping

kuujo:
I think I don’t understand this change. If `eventListeners` is empty then `eventListeners.forEach(...)` is effectively a no-op. What am I missing?

johnou:
@zwillim please see the above questions and respond when you have time, I will close the pull request in the meantime.

PR #845, opened 2018-09-26
Support binding to local interface distinct from broadcast address

kuujo:
This PR adds support for binding to interfaces that are distinct from the ones reachable by other members of the cluster. This is useful, for instance, in containerized environments where the node binds to a private container IP/port but is only reachable through the host IP/port.


kuujo:
I changed the configuration to add a `bind` object for the local interfaces/port:

```
cluster.bind {
  interfaces: ["127.0.0.1", "123.123.123.123"]
  port: 5000
}
```

kuujo:
@johnou I can't see your review comments now 🤷‍♂️

kuujo:
Bah it's because I squashed the changes. Bad habit from Gerrit reviews...

johnou:
No changes required for messaging or timeout detection?

PR #846, opened 2018-09-26
Listen on all interfaces when running Atomix agent

kuujo:
There's currently no way to specify an interface for the HTTP server to listen on, so this PR ensures the server listens on all available interfaces.


kuujo:
I think it's pretty common to default to `0.0.0.0`. Really, in terms of security it could/should be managed at the network layer. But more importantly we do need to at least provide a way to override this default via CLI options. Currently only the port can be set.

kuujo:
This PR got messed up somehow...

PR #847, opened 2018-09-26
Throw protocol exception to close connection on invalid version

kuujo:
This PR improves support for changes to the messaging protocol by adding an exception when a protocol version mismatch is detected. This will be needed to support downgrading protocol versions on failed handshakes.


cluster/src/main/java/io/atomix/cluster/messaging/impl/MessageDecoder.java
```diff
@@ -70,6 +72,9 @@ protected void decode(
           return;
         }
         version = buffer.readShort();
+        if (version != VERSION) {
+          throw new ProtocolException("Unsupported protocol version: " + version);
```
johnou: I think before throwing the exception we should be releasing the bytebuf.

PR #850, opened 2018-09-27
Support parsing strings for numeric values in ConfigMapper

kuujo:
This PR adds support for parsing numbers from strings e.g. in JSON configurations:

```
{
  "foo": "1"
}
```

The value `"1"` can be applied to a method that takes an `int` argument.


johnou:
nice!

PR #853, opened 2018-10-01
Ensure cached document tree properly registers root listener

kuujo:
This PR fixes a bug in the cached `DocumentTree` primitive listeners. The primitive registers a root listener to ensure the cache is updated before events are received by the client. However, the root listener is not currently registered correctly. It currently just adds a listener to itself. This PR ensures the listener is registered on the underlying proxy using the `root()` path.


kuujo:
@Andrea-Campanella

PR #863, opened 2018-10-03
Select new primary from existing backups

alexrobin:


primitive/src/main/java/io/atomix/primitive/partition/impl/PrimaryElectorService.java
```diff
@@ -310,10 +311,19 @@ ElectionState cleanup(Session session) {
                 .collect(Collectors.toList());
         if (primary.sessionId() == session.sessionId().id()) {
           if (!updatedRegistrations.isEmpty()) {
+            // find 1st backup to use as primary
+            Registration newPrimary = updatedRegistrations.get(0);
+            List<GroupMember> backups = term().backups(1);
+            if (!backups.isEmpty())
+            {
```
kuujo: Code style - brackets on same line as condition

primitive/src/main/java/io/atomix/primitive/partition/impl/PrimaryElectorService.java
```diff
@@ -103,6 +103,7 @@ protected void configure(ServiceExecutor executor) {
   }
 
   private void notifyTermChange(PartitionId partitionId, PrimaryTerm term) {
+    getLogger().debug("Changed " + term + ", listeners=" + listeners.size());
```
kuujo: We should always use string replacement with `{}`

primitive/src/main/java/io/atomix/primitive/partition/impl/PrimaryElectorService.java
```diff
@@ -310,10 +311,19 @@ ElectionState cleanup(Session session) {
                 .collect(Collectors.toList());
         if (primary.sessionId() == session.sessionId().id()) {
           if (!updatedRegistrations.isEmpty()) {
+            // find 1st backup to use as primary
+            Registration newPrimary = updatedRegistrations.get(0);
+            List<GroupMember> backups = term().backups(1);
```
kuujo: This change doesn’t make sense to me. The registrations are already sorted such that the primary is `registrations.get(0)` and the backups are `registrations.subList(1, numBackups)`. If we remove the primary (`registrations.remove(0)`) then the next primary (`registrations.get(0)`) is already a backup.

alexrobin: The issue comes up when using a strategy like `MemberGroupStrategy.RACK_AWARE` for backup selection. In this case, the selected backups are not always `registrations.subList(1, numBackups)`, because the `PrimaryTerm.backups(int)` method may return different ones to avoid selecting two in the same group.

primitive/src/main/java/io/atomix/primitive/partition/impl/PrimaryElectorService.java
alexrobin: yes, sorry, I will fix this and update the PR

primitive/src/main/java/io/atomix/primitive/partition/impl/PrimaryElectorService.java
alexrobin: Ok, will fix and update the PR

primitive/src/main/java/io/atomix/primitive/partition/impl/PrimaryElectorService.java
kuujo: I gotcha. This is indeed problematic. So the primary could be in e.g. group `1` and then fail over to a primary that's not actually used as a backup, thus potentially causing the backups to be recomputed as well if the new primary is in a different group. I think I need to think about this a bit...

kuujo:
I feel like we should be sorting the candidates by group in the primary elector service rather than sorting them FIFO and selecting them in the client side. That would give us much better control over how data is being replicated.

alexrobin:
Yes, or maybe the primary elector should also be the one selecting the actual backups, rather than just a list of candidates? This way all the decisions are made in a central place. If we let the client make this decision, there is always a risk that the elector thinks it was a backup but it wasn't, which can cause it to elect a suboptimal primary.

kuujo:
Well, the problem is each partition shares a primary election, but each primitive can be configured with n backups. So we need to give the client a prioritized list of all candidates, and it should just take the primary + n backups and the service should sort the candidates by group.

kuujo:
Yes, please do.

kuujo:
Closed in favor of #991

Issue #868, opened 2018-10-05
NodeDiscoveryProvider protocols are redundant


kuujo:
The current implementations of `NodeDiscoveryProvider`s are redundant. They use phi accrual failure detectors to add/remove nodes from the cluster. This should be the purview of the cluster membership protocols, not the providers. The providers should be responsible only for notifying the membership protocol of new members to probe. Refactor existing providers and the membership service to manage failure detection outside of providers.

Note that this needs to be done in a compatible manner to preserve rolling upgrades including 3.0.x nodes in which providers detect failures. This can be done by moving provider heartbeats into the membership service.

kuujo:
Fixed in 3.1

PR #870, opened 2018-10-05
Simplify node discovery providers

kuujo:
This PR refactors node discovery providers to move failure detection solely under the purview of `ClusterMembershipService` per #868

### DIFF
--- a/cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMembershipService.java
+++ b/cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMembershipService.java
@@ -16,6 +16,7 @@
 package io.atomix.cluster.impl;
 
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import io.atomix.cluster.BootstrapService;
 import io.atomix.cluster.ClusterMembershipEvent;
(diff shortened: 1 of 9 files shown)


cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMembershipService.java
```diff
@@ -152,118 +157,138 @@ private void handleDiscoveryEvent(NodeDiscoveryEvent event) {
...
+      members.put(remoteMember.id(), remoteMember);
+      post(new ClusterMembershipEvent(ClusterMembershipEvent.Type.MEMBER_ADDED, remoteMember));
+    } else {
+      // If the member's metadata changed, overwrite the local member. We do this instead of a version check since
+      // a heartbeat from the source member always takes precedence over a more recent version.
+      if (!Objects.equals(member.version(), remoteMember.version())
+          || !Objects.equals(member.zone(), remoteMember.zone())
+          || !Objects.equals(member.rack(), remoteMember.rack())
+          || !Objects.equals(member.host(), remoteMember.host())
+          || !Objects.equals(member.properties(), remoteMember.properties())) {
```
johnou: forget which type properties is, should we be using Arrays.equals here or Objects.equals does the job?

utils/src/main/java/io/atomix/utils/config/ConfigMapper.java
```diff
@@ -558,20 +573,24 @@ protected static String toSetterName(String name) {
   protected static class SetterDescriptor {
     private final String name;
     private final Method setter;
+    private final boolean deprecated;
 
     SetterDescriptor(String name, Method setter) {
       this.name = name;
       this.setter = setter;
+      this.deprecated = setter.getAnnotation(Deprecated.class) != null;
```
johnou: nice!

kuujo: I know, I really like that feature. We need to log a lot better information about configuration options. Another nice feature would be making suggestions for invalid properties, e.g. using Levenshtein distance to match a real property or something. `Did you mean 'cluster.members.1.id'?`

cluster/src/main/java/io/atomix/cluster/impl/DefaultClusterMembershipService.java
kuujo: It's a `Properties` object

Issue #866, opened 2018-10-05
Pluggable cluster membership protocol


kuujo:
The current cluster membership protocol is is a gossip-based protocol that uses phi actual failure detectors. This protocol is decent for small clusters, but will perform poorly in large clusters. The protocol can and should be abstracted from the `ClusterMembershipService` to allow users to configure the protocol to be used for group membership.

kuujo:
I think we're still missing `Atomix` builder methods for programmatic configuration.

PR #871, opened 2018-10-06
Pluggable membership protocol

kuujo:
This PR implements a pluggable group membership protocol as per #866

The protocol is configured via the `cluster.protocol` object, and the configuration has the same basic format as all other extensible configurations:
```
cluster.protocol {
  type: phi
  heartbeatInterval: 100ms
}
```

The default protocol is `phi`, which is the membership protocol which has been used in Atomix since 3.0.

### DIFF
--- a/cluster/src/main/java/io/atomix/cluster/AtomixClusterBuilder.java
+++ b/cluster/src/main/java/io/atomix/cluster/AtomixClusterBuilder.java
@@ -17,6 +17,8 @@ package io.atomix.cluster;
 
 import com.google.common.collect.Lists;
 import io.atomix.cluster.discovery.NodeDiscoveryProvider;
+import io.atomix.cluster.protocol.PhiMembershipProtocolConfig;
+import io.atomix.cluster.protocol.GroupMembershipProtocolConfig;
 import io.atomix.utils.Builder;
 import io.atomix.utils.net.Address;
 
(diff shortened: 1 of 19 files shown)


cluster/src/main/java/io/atomix/cluster/AtomixClusterBuilder.java
```diff
@@ -328,7 +330,10 @@ public AtomixClusterBuilder withMulticastAddress(Address address) {
    */
   @Deprecated
   public AtomixClusterBuilder setBroadcastInterval(Duration interval) {
-    config.getMembershipConfig().setBroadcastInterval(interval);
+    GroupMembershipProtocolConfig protocolConfig = config.getProtocolConfig();
+    if (protocolConfig instanceof PhiMembershipProtocolConfig) {
+      ((PhiMembershipProtocolConfig) protocolConfig).setHeartbeatInterval(interval);
```
johnou: might be a bit cleaner for GroupMembershipProtocolConfig to implement a generic interface method setBroadcastInterval?

cluster/src/main/java/io/atomix/cluster/AtomixClusterBuilder.java
```diff
@@ -357,7 +366,10 @@ public AtomixClusterBuilder withBroadcastInterval(Duration interval) {
    */
   @Deprecated
   public AtomixClusterBuilder setReachabilityThreshold(int threshold) {
-    config.getMembershipConfig().setReachabilityThreshold(threshold);
+    GroupMembershipProtocolConfig protocolConfig = config.getProtocolConfig();
+    if (protocolConfig instanceof PhiMembershipProtocolConfig) {
+      ((PhiMembershipProtocolConfig) protocolConfig).setFailureThreshold(threshold);
```
johnou: and for this

cluster/src/main/java/io/atomix/cluster/AtomixClusterBuilder.java
kuujo: This is just for backwards compatibility

cluster/src/main/java/io/atomix/cluster/AtomixClusterBuilder.java
kuujo: ditto

PR #872, opened 2018-10-06
DNS/SRV discovery provider

kuujo:
This PR introduces a `NodeDiscoveryProvider` for DNS using SRV records. This provider can be used in Kubernetes to discover the host/port of Atomix nodes.

This is a work in progress. Need to merge earlier PRs and test in k8s before this can be reviewed and merged.


cluster/src/main/java/io/atomix/cluster/discovery/DnsDiscoveryProvider.java
```diff
@@ -0,0 +1,172 @@
...
+
+  public static final Type TYPE = new Type();
+
+  /**
+   * Creates a new bootstrap provider builder.
+   *
+   * @return a new bootstrap provider builder
+   */
+  public static BootstrapDiscoveryBuilder builder() {
+    return new BootstrapDiscoveryBuilder();
```
kuujo: Returning the wrong provider

Issue #864, opened 2018-10-05
Messaging handshake protocol


kuujo:
In order to support backward/forward compatible changes to the message protocol for rolling upgrades, a new protocol needs to be developed which negotiates the protocol version. The new protocol should include a handshake protocol to determine the version, and use the appropriate encoder/decoder for the protocol version.

Additionally, the new protocol should be capable of detecting invalid packets and closing connections. It should also limit the amount of memory that will be allocated for messages to prevent DOS attacks.

PR #873, opened 2018-10-10
Messaging handshake protocol

kuujo:
This PR implements a new handshake protocol for the `NettyMessagingService` as per #864

The handshake protocol works by sending a six byte hello when a client connects to a server. The six bytes consist of the 4 byte preamble and 2 byte protocol version number. The preamble is used to verify the connection is from another Atomix node. The server returns the highest compatible protocol version, and the `ProtocolVersion` is used to add an encoder/decoder to the channel handler pipeline on both the client and server for the negotiated version of the protocol.


cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyMessagingService.java
```diff
@@ -198,10 +201,10 @@ public ManagedMessagingService build() {
   private final Integer bindPort;
   private final Address returnAddress;
   private final AtomicBoolean started = new AtomicBoolean(false);
-  private final Map<String, BiConsumer<InternalRequest, ServerConnection>> handlers = new ConcurrentHashMap<>();
+  private final Map<String, BiConsumer<ProtocolRequest, ServerConnection>> handlers = new ConcurrentHashMap<>();
   private final Map<Channel, RemoteClientConnection> clientConnections = Maps.newConcurrentMap();
   private final Map<Channel, RemoteServerConnection> serverConnections = Maps.newConcurrentMap();
-  private final AtomicLong messageIdGenerator = new AtomicLong(0);
+  private final AtomicInteger messageIdGenerator = new AtomicInteger(0);
```
kuujo: This should be an `AtomicLong`

cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyMessagingService.java
```diff
@@ -752,49 +768,146 @@ protected void initChannel(SocketChannel channel) throws Exception {
...
+    @Override
+    public void channelRead(ChannelHandlerContext context, Object message) throws Exception {
+      readProtocolVersion(context, (ByteBuf) message)
+          .ifPresent(version -> activateProtocolVersion(context, version));
+    }
+
+    @Override
+    void activateProtocolVersion(ChannelHandlerContext context, ProtocolVersion protocolVersion) {
+      super.activateProtocolVersion(context, protocolVersion);
+      future.complete(context.channel());
```
kuujo: The future needs to be completed exceptionally if an exception occurs in the pipeline.

PR #878, opened 2018-10-11
Detect multiple metadata changes on local node

kuujo:
This PR fixes a bug in detecting local metadata changes and triggering events. Because the `DefaultClusterMembershipService` stores a reference to the updated metadata object, future equality checks are unsuccessful even if metadata has changed.


kuujo:
This change likely needs to be cherry picked into the 3.1 branch as well.

PR #884, opened 2018-10-12
Ensure transaction participants are closed on commit

kuujo:
This PR fixes a bug in transactions. Currently, transaction participants' sessions are not closed when a transaction is committed.

### DIFF
--- a/core/src/main/java/io/atomix/core/transaction/impl/DefaultTransaction.java
+++ b/core/src/main/java/io/atomix/core/transaction/impl/DefaultTransaction.java
@@ -127,7 +127,9 @@ public class DefaultTransaction implements AsyncTransaction {
             : transactionService.aborting(transactionId)
             .thenCompose(v -> rollback(participants))
             .thenApply(v -> CommitStatus.FAILURE));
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/atomix/core/transaction/impl/DefaultTransaction.java
```diff
@@ -127,7 +127,9 @@ void addParticipants(Collection<TransactionParticipant<?>> participants) {
             : transactionService.aborting(transactionId)
             .thenCompose(v -> rollback(participants))
             .thenApply(v -> CommitStatus.FAILURE));
-    return status.thenCompose(v -> transactionService.complete(transactionId).thenApply(u -> v));
+    return status.thenCompose(v -> transactionService.complete(transactionId)
+        .thenRun(() -> participants.forEach(TransactionParticipant::close))
```
johnou: is it possible close may thow an exception and fail the future?

kuujo: I think it could indeed

Issue #885, opened 2018-10-13
Add Singleton get method to primitive builders


kuujo:
The current mechanism for singleton primitive instances is the `get*` methods on the `Atomix` interface. But these methods do not allow for programmatic configuration of singleton primitives. To allow programmatic configuration, we should add `get()` and `getAsync()` methods to `PrimitiveBuilder`. These methods would use the singleton primitive cache that’s currently used by the `Atomix` singletons.

PR #886, opened 2018-10-14
Add getter methods to primitive builders for singletons.

kuujo:
#885

Adds singleton getters to primitive builders and deprecates `get*` methods on the `Atomix` API. The singleton methods added to `PrimitiveBuilder` are:
* `get()` - synchronous singleton getter
* `getAsync()` - asynchronous singleton getter


kuujo:
rebase

PR #892, opened 2018-10-18
Ensure unknown members are propagated in phi accrual protocol responses

kuujo:
This PR fixes a bug in the phi membership protocol that prevents unknown members (not provided by the discovery provider) from being propagated. It adds full cluster membership information to the phi protocol responses.

### DIFF
--- a/cluster/src/main/java/io/atomix/cluster/protocol/PhiMembershipProtocol.java
+++ b/cluster/src/main/java/io/atomix/cluster/protocol/PhiMembershipProtocol.java
@@ -16,6 +16,7 @@
 package io.atomix.cluster.protocol;
 
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Lists;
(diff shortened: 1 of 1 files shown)


cluster/src/main/java/io/atomix/cluster/protocol/PhiMembershipProtocol.java
```diff
@@ -287,8 +297,13 @@ private void updateMember(GossipMember remoteMember) {
       }
       localMember.properties().putAll(remoteMember.properties());
       post(new GroupMembershipEvent(GroupMembershipEvent.Type.METADATA_CHANGED, localMember));
-    } else if (!localMember.isReachable()) {
+    } else if (!localMember.isReachable() && direct) {
+      localMember.setReachable(true);
```
johnou: should we be checking remoteMember reachable here? is it possible for a member to receive two not reachable updates?

PR #896, opened 2018-10-26
Implement file channel based journal for Raft log

kuujo:
This PR replaces the existing journal with a new, more efficient `FileChannel` based journal. The `FileChannel` journal uses `Namespace`s to serialize/deserialize objects from the journal, thus avoiding an unnecessary memory copy when using memory mapped files. This PR also deprecates the use of the `MEMORY` storage level.


storage/src/main/java/io/atomix/storage/journal/FileChannelJournalSegmentReader.java
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  @Override
+  public Indexed<E> getCurrentEntry() {
+    return currentEntry;
+  }
+
+  @Override
+  public long getNextIndex() {
+    return currentEntry != null ? currentEntry.index() + 1 : firstIndex;
```
johnou: are entries guaranteed to be contiguous? why aren't we using nextEntry here?

storage/src/main/java/io/atomix/storage/journal/FileChannelJournalSegmentWriter.java
```diff
@@ -0,0 +1,333 @@
...
+    lastEntry = null;
+
+    try {
+      // Truncate the index.
+      this.index.truncate(index);
+
+      if (index < segment.index()) {
+        channel.position(JournalSegmentDescriptor.BYTES);
+        memory.clear();
+        for (int i = 0; i < memory.limit(); i++) {
```
johnou: utility method?

storage/src/main/java/io/atomix/storage/journal/SegmentedJournal.java
```diff
@@ -369,115 +391,82 @@ synchronized void removeSegment(JournalSegment segment) {
...
-    Buffer buffer = HeapBuffer.allocate(descriptor.maxSegmentSize(), descriptor.maxSegmentSize());
+    ByteBuffer buffer = ByteBuffer.allocate(JournalSegmentDescriptor.BYTES);
     descriptor.copyTo(buffer);
+    buffer.flip();
+    try {
+      channel.write(buffer);
+    } catch (IOException e) {
+      throw new StorageException(e);
+    } finally {
+      try {
```
johnou: release `buffer` too?

kuujo: It's a `ByteBuffer`

storage/src/main/java/io/atomix/storage/journal/FileChannelJournalSegmentReader.java
kuujo: Yes they're guaranteed to be contiguous. `nextEntry` gets its index from `getNextIndex()`, so I think there might be some weird problems with using it.

Issue #794, opened 2018-08-09
Expose Netty TLS configuration via builders


kuujo:
The Netty TLS configuration is currently setup via system properties. These options should be exposed in configuration files and builders.

PR #901, opened 2018-10-26
Use Netty SslContextBuilder for configuring TLS.

johnou:
Part of https://github.com/atomix/atomix/issues/794


kuujo:
I think we need to add an option to enable SSL in the test framework for verification

PR #902, opened 2018-10-29
Fix race condition in cluster membership test

kuujo:
This PR fixes a bug in the `ClusterMembershipService` tests as well as in the `ClusterMembershipService` itself. The bug in the test was a race condition that could result in incorrect events being asserted. The bug in the service itself was also a race condition that could prevent metadata changes from being detected when performed closely together.


johnou:
Race condition where it would update metadata while updating metadata?

kuujo:
Indeed

johnou:
Nice catch.

Issue #903, opened 2018-10-30
Add distributed log


kuujo:
Atomix has always been moving towards a distributed log primitive simile to Kafka. Indeed, this is the very reason the storage abstractions a separate, independent module. We’ve now encountered a need for a distributed log and so need to implement the distributed log primitive.

My plan is to add a `DistributedLogPartitonGroup` that provides distributed log partitions. Distributed log partitions work much the same way as the primary-backup protocol, using the primary election API to elect a leader/followers for each partition and replicate sync/async based on configuration. The log will be a first class primitive abstraction in the `io.atomix.primitive.log` package just like proxies are a first class abstraction. This means there will be two types of protocols: proxy and log.

Similarly, a `DistributedLogProtocol` will be used to configure a `DistributedLog`/`AsyncDistributedLog` primitive.

An interesting feature of the `DistributedLogProtocol` is that it can also be used on any state-machine-based primitive. When the `DistributedLogProtocol` is used on e.g. a map, the map service will be constructed on the client side rather than the server side and the distributed log will be read to populate the map service, then map reads are local. This justifies logs as first class primitive protocols.

The question is what the distributed log API should look like. There are several different approaches that can be taken. I’m leaving that up for debate once the implementation is complete. But at the very least, the log must support batch reads/writes as well as streaming. Exactly-once semantics can be left for a later issue.

PR #905, opened 2018-11-01
[WIP] Add distributed log protocol/primitive

kuujo:
This PR is an initial implementation of a distributed log protocol and primitive re: #903 

The log protocol is implemented by adding a new core primitive API for logs: `LogClient`, `LogSession`, and `LogRecord`. The implementation of these interfaces is the distributed `log` protocol. As with other protocols, a `LogPartitionGroup`/`LogPartition` is used to configure a distributed log, and a `DistributedLogProtocol` is used to configure primitives to use the distributed log. The `DistributedLog` primitive is the only primitive that is not a state machine that supports the `DistributedLogProtocol`.

All other primitives can use the `DistributedLogProtocol` as well. When the `DistributedLogProtocol` is used on a primitive, the `LogProxySession` implements the `ProxySession` for the primitive and handles writing primitive operations to the distributed log and applying operations to a local `PrimitiveService` when reading from the distributed log. Reads are performed locally in a sequentially consistent manner.

The distributed log protocol itself works much like the primary-backup protocol. It uses primary elections to elect a leader and followers and replicates logs from the leader to the followers. The leader is responsible for all client requests. Client consumers receive distributed log records from the leader over UDP to reduce overhead, and missing log records are resolved by tracking indexes on the client to ensure records are received in sequential order. When the network drops a log record, the client requests records starting at the last received index from the leader.

Logs are persisted to disk on all participating nodes using a `SegmentedJournal`. Nodes support size or time-based compaction of the distributed log, retaining as much information as possible as opposed to e.g. Raft which tends to compact logs sooner.

The `DistributedLog` primitive does not track and store consumer offsets. Instead, users can use other primitives (e.g. `AtomicCounter` or `AtomicCounterMap`) to track consumer offsets, and the `DistributedLog` API provides a mechanism for consumers to start at an arbitrary offset.

Distributed log primitives are not named like other primitives since a `DistributedLog` primitive reads all entries in a distributed log partition group.

```java
DistributedLog<String> log = atomix.<String>logBuilder()
  .withProtocol(DistributedLogProtocol.builder()
    ...
    .build())
  .build();

log.produce("Hello world!");
log.consume(record -> {
  long offset = record.offset();
  long timestamp = record.timestamp();
  String value = record.value();
});
```


kuujo:
BTW there were a few bugs fixed in this PR relating to primary elections and cluster communication. I will submit them separately as PRs to the 3.0 branch as well.

core/src/main/java/io/atomix/core/log/impl/DefaultAsyncDistributedLog.java
```diff
@@ -0,0 +1,136 @@
...
+    client.getPartitions().forEach(partition -> {
+      DefaultAsyncDistributedLogPartition<E> logPartition = new DefaultAsyncDistributedLogPartition<>(this, partition, serializer);
+      partitions.put(partition.partitionId().id(), logPartition);
+      sortedPartitions.add(logPartition);
+    });
+  }
+
+  @Override
+  public String name() {
+    return null;
```
johnou: shouldn't this be assigned?

primitive/src/main/java/io/atomix/primitive/session/impl/AbstractSession.java
```diff
@@ -47,7 +47,7 @@ protected AbstractSession(
     this.sessionId = checkNotNull(sessionId);
     this.primitiveName = checkNotNull(primitiveName);
     this.primitiveType = checkNotNull(primitiveType);
-    this.memberId = checkNotNull(memberId);
+    this.memberId = memberId;
```
johnou: was this an accident?

protocols/log/src/main/java/io/atomix/protocols/log/serializer/impl/LogNamespaces.java
```diff
@@ -0,0 +1,62 @@
...
+      .register(ConsumeRequest.class)
+      .register(ConsumeResponse.class)
+      .register(RecordsRequest.class)
+      .register(ResetRequest.class)
+      .register(BackupRequest.class)
+      .register(BackupResponse.class)
+      .register(BackupOperation.class)
+      .register(LogEntry.class)
+      .register(LogRecord.class)
+      .build("PrimaryBackupProtocol");
```
johnou: shouldn't this read LogProtocol?

protocols/log/src/main/java/io/atomix/protocols/log/impl/DistributedLogSession.java
```diff
@@ -0,0 +1,295 @@
...
+  private final ThreadContext threadContext;
+  private final Set<Consumer<PrimitiveState>> stateChangeListeners = Sets.newIdentityHashSet();
+  private final ClusterMembershipEventListener membershipEventListener = this::handleClusterEvent;
+  private final PrimaryElectionEventListener primaryElectionListener = event -> changeReplicas(event.term());
+  private final DistributedLogProducer producer = new DistributedLogProducer();
+  private final DistributedLogConsumer consumer = new DistributedLogConsumer();
+  private final MemberId memberId;
+  private final String subject;
+  private PrimaryTerm term;
+  private volatile PrimitiveState state = PrimitiveState.CONNECTED;
```
johnou: shouldn't the init state be closed?

protocols/log/src/main/java/io/atomix/protocols/log/DistributedLogSessionClient.java
```diff
@@ -0,0 +1,261 @@
...
+    public DistributedLogSessionClient build() {
+      Logger log = ContextualLoggerFactory.getLogger(DistributedLogSessionClient.class, LoggerContext.builder(DistributedLogSessionClient.class)
+          .addValue(clientName)
+          .build());
+
+      // If a ThreadContextFactory was not provided, create one and ensure it's closed when the client is stopped.
+      boolean closeOnStop;
+      ThreadContextFactory threadContextFactory;
+      if (this.threadContextFactory == null) {
+        threadContextFactory = threadModel.factory("backup-client-" + clientName + "-%d", threadPoolSize, log);
```
kuujo: The thread name needs to be changed

primitive/src/main/java/io/atomix/primitive/session/impl/AbstractSession.java
kuujo: No it's intentional. There's no member information with distributed log entries, so the sessions just set it to `null`

protocols/log/src/main/java/io/atomix/protocols/log/roles/LogServerRole.java
```diff
@@ -0,0 +1,133 @@
...
+   * @return the role type
+   */
+  public Role role() {
+    return role;
+  }
+
+  /**
+   * Logs a request.
+   */
+  protected final <R extends LogRequest> R logRequest(R request) {
```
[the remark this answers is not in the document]
kuujo: Received request, sent response

PR #912, opened 2018-11-07
Fix incorrect SWIM protocol name for file-based configurations

kuujo:


kuujo:
That’s why this branch is still unreleased 🤷‍♂️

PR #916, opened 2018-11-09
Retry primitive operations on SocketException

kuujo:


protocols/raft/src/main/java/io/atomix/protocols/raft/session/impl/RaftSessionConnection.java
```diff
@@ -267,7 +268,7 @@ public MemberId leader() {
         error = error.getCause();
       }
       log.debug("{} failed! Reason: {}", request, error);
-      if (error instanceof ConnectException || error instanceof TimeoutException || error instanceof ClosedChannelException) {
+      if (error instanceof SocketException || error instanceof TimeoutException || error instanceof ClosedChannelException) {
```
johnou: and ConnectException?

kuujo: `ConnectException` extends `SocketException`

PR #920, opened 2018-11-09
Use max bytes receive allocator in Netty unicast service

kuujo:


kuujo:
Actually it just fixes a bug with sending large packets. I haven't yet tested the throughput.

johnou:
What bug? Can you reproduce it?

kuujo:
If a packet larger than 2048 bytes is sent then it's not aggregated on the
receiving side without this change.

PR #922, opened 2018-11-12
Replace WARN level configuration messages with DEBUG

kuujo:
Many users have asked what the warnings mean, and they're typically benign unless repeating. The configuration protocol is stable enough that they can be changed to DEBUG level messages.


kuujo:
This will need to be cherry picked to 3.0 and 2.0 branches.

PR #924, opened 2018-11-12
Fix improper formatting of deprecation warnings

kuujo:
Fixes a bug in logging of configuration deprecation warnings, handling null paths.


kuujo:
Needs to be cherry picked to 3.0.

PR #929, opened 2018-11-13
SWIM protocol improvements

kuujo:
This PR adds several improvements to the SWIM protocol implementation:
- Replicate local member info on probes
- Subtract local member from probe request count
- Set configurable probe timeout
- Treat probe request timeouts as failures

### DIFF
--- a/cluster/src/main/java/io/atomix/cluster/protocol/SwimMembershipProtocol.java
+++ b/cluster/src/main/java/io/atomix/cluster/protocol/SwimMembershipProtocol.java
@@ -32,6 +32,8 @@ import io.atomix.utils.net.Address;
 import io.atomix.utils.serializer.Namespace;
 import io.atomix.utils.serializer.Namespaces;
 import io.atomix.utils.serializer.Serializer;
+import org.apache.commons.lang3.tuple.ImmutablePair;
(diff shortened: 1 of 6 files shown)


cluster/src/main/java/io/atomix/cluster/protocol/SwimMembershipProtocol.java
```diff
@@ -380,7 +383,8 @@ private void probe(boolean all) {
    */
   private void probe(ImmutableMember member) {
     LOGGER.trace("{} - Probing {}", localMember.id(), member);
-    bootstrapService.getMessagingService().sendAndReceive(member.address(), MEMBERSHIP_PROBE, SERIALIZER.encode(member))
+    bootstrapService.getMessagingService().sendAndReceive(
+        member.address(), MEMBERSHIP_PROBE, SERIALIZER.encode(Pair.of(localMember.copy(), member)), config.getProbeTimeout())
```
johnou: should we be taking a copy of the member too?

kuujo: The one passed in to the method is copied already. Basically, there's a mutable member and immutable member. The mutable member's `copy()` method just returns an immutable member which is used for replication.

PR #936, opened 2018-11-17
Refactor NettyMessagingService

kuujo:
This PR refactors the `NettyMessagingService` to clean up the implementation and reduce the overhead of dynamic timeouts. 

* Many of the classes that were previously internal to the `NettyMessagingService` are now moved to package private classes since the class was getting a bit unwieldy
* The channel pool is moved to a separate `ChannelPool` class
* Local and remote client and server connections share a base interface and class through which they share a lot of logic
* Compute dynamic message timeouts before request to allow timeout scheduling at the start of the request. This reduces the currently significant overhead from constantly recomputing timeouts while messages are in flight.
* Use `ScheduledExecutorService` directly for static and dynamic message timeouts
* Use a simple max-based algorithm for computing dynamic timeouts

### DIFF
--- a/cluster/src/main/java/io/atomix/cluster/messaging/impl/AbstractClientConnection.java
+++ b/cluster/src/main/java/io/atomix/cluster/messaging/impl/AbstractClientConnection.java
@@ -1,0 +1,184 @@
+/*
+ * Copyright 2018-present Open Networking Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.atomix.cluster.messaging.impl;
+
+import com.google.common.collect.Maps;
+import io.atomix.cluster.messaging.MessagingException;
+import org.apache.commons.math3.stat.descriptive.DescriptiveStatistics;
+import org.apache.commons.math3.stat.descriptive.SynchronizedDescriptiveStatistics;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.ConnectException;
+import java.time.Duration;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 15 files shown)


cluster/src/main/java/io/atomix/cluster/messaging/impl/LocalClientConnection.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Local client-side connection.
+ */
+final class LocalClientConnection extends AbstractClientConnection {
+  private final LocalServerConnection serverConnection;
+
+  LocalClientConnection(ScheduledExecutorService executorService, HandlerRegistry handlers) {
+    super(executorService);
+    this.serverConnection = new LocalServerConnection(handlers);
+    serverConnection.setClientConnection(this);
```
johnou: Could this be set in the constructor?

kuujo: Yes good point.

cluster/src/main/java/io/atomix/cluster/messaging/impl/AbstractClientConnection.java
```diff
@@ -0,0 +1,183 @@
...
+     */
+    String type() {
+      return type;
+    }
+
+    /**
+     * Fails the callback future with a timeout exception.
+     */
+    private void timeout() {
+      replyFuture.completeExceptionally(new TimeoutException("Request type " + type + " timed out in " + timeout + " milliseconds"));
```
kuujo: I think this method may need to remove the callback from `callbacks` here

kuujo:
Need to profile this change to ensure there are no memory leaks before merging.

cluster/src/main/java/io/atomix/cluster/messaging/impl/AbstractClientConnection.java
johnou: Yeah or risk of memory leak?

cluster/src/main/java/io/atomix/cluster/messaging/impl/LocalClientConnection.java
```diff
@@ -0,0 +1,52 @@
...
+  public CompletableFuture<byte[]> sendAndReceive(ProtocolRequest message, Duration timeout) {
+    CompletableFuture<byte[]> future = new CompletableFuture<>();
+    new Callback(message.id(), message.subject(), timeout, future);
+    serverConnection.dispatch(message);
+    return future;
+  }
+
+  @Override
+  public void close() {
+    super.close();
```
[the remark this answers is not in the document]
kuujo: Nope... all it really does is ensure that pending futures are completed with a `ConnectException`

PR #949, opened 2018-12-10
Benchmark implementation

kuujo:
This PR adds a new module for benchmarking Atomix clusters. The benchmark tests replication protocols specifically. The way benchmarks work is by using a special add-on nodes to execute and benchmark `AtomicMap` operations against an arbitrary partition group. To run a benchmark, you simply add some arbitrary number of benchmark nodes by using the `atomix-bench` script. 

```
./bin/atomix-bench -c bench.conf
```

More bench nodes effectively tests more clients/larger clusters. Once the benchmark nodes have been set up, run the `python3 run_bench.py` script with arguments specifying the percentage of reads and writes, number of operations, concurrency within each node, key/value sizes, whether to benchmark events, etc.

```
python3 ./bin/run_bench.py --host localhost --port 5678 --protocol multi-raft --group raft --ops 10000 --concurrency 10
```

When the benchmark is run, the bench node through which the test is run (specified by `--host` and `--port` in `run_bench.py`) will coordinate the benchmark, assigning a subset of the operations to each of the available `atomix-bench` nodes. Those nodes will periodically report their progress (once a second) and then send a final report to the coordinator once all their respective operations are completed. The script then polls the benchmark REST API to receive progress reports and print the final benchmark result.

There's still a lot of work to do on the actual numbers that are recorded. Currently the script more or less just prints each process' throughput. We should also be recording the latency of reads and writes and report quantiles.

The `atomix-test` framework will be updated separately to support easily forming clusters for benchmarking and running benchmarks.

### DIFF
--- a/bench/src/main/java/io/atomix/bench/AtomixBenchmark.java
+++ b/bench/src/main/java/io/atomix/bench/AtomixBenchmark.java
@@ -71,7 +71,7 @@ public class AtomixBenchmark {
     LOGGER.warn("The Atomix HTTP API is BETA and is intended for development and debugging purposes only!");
     LOGGER.info("HTTP server listening at {}", rest.address());
 
-    new AtomixBenchmark(buildAtomix(namespace)).start();
+    new AtomixBenchmark(atomix).start();
 
     synchronized (Atomix.class) {
       while (atomix.isRunning()) {
--- a/bench/src/main/java/io/atomix/bench/AtomixBenchmark.java
+++ b/bench/src/main/java/io/atomix/bench/AtomixBenchmark.java
@@ -22,6 +22,8 @@ import io.atomix.cluster.discovery.BootstrapDiscoveryConfig;
 import io.atomix.cluster.discovery.MulticastDiscoveryConfig;
 import io.atomix.core.Atomix;
 import io.atomix.core.AtomixConfig;
+import io.atomix.rest.ManagedRestService;
+import io.atomix.rest.RestService;
 import io.atomix.utils.net.Address;
 import io.atomix.utils.net.MalformedAddressException;
 import net.sourceforge.argparse4j.ArgumentParsers;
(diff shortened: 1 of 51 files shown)


natraj09:
@kuujo I am interested in benchmarking AtomixQueue , what additional changes are required , any pointer?

bench/src/main/java/io/atomix/bench/AtomixBenchmark.java
```diff
@@ -62,25 +62,37 @@ public static void main(String[] args) throws Exception {
...
     extraArgs.getAttrs().forEach((key, value) -> System.setProperty(key, value.toString()));
-    new AtomixBenchmark(namespace).start();
+
+    Atomix atomix = buildAtomix(namespace);
+    atomix.start().join();
+    LOGGER.info("Atomix listening at {}", atomix.getMembershipService().getLocalMember().address());
+
+    final ManagedRestService rest = buildRestService(atomix, namespace);
+    rest.start().join();
+    LOGGER.warn("The Atomix HTTP API is BETA and is intended for development and debugging purposes only!");
```
johnou: why in the benchmark?

kuujo:
@natraj09 I updated the framework to support different types of benchmarks. So, a benchmark is created by doing `POST the-server:the-port/v1/bench` with a configuration that can now specify the benchmark type. So, you would do something like implement a `QueueBenchmarkController` and `QueueBenchmarkExecutor` and add a `QUEUE` type to `BenchmarkType` and define a benchmark configuration, then you could post a queue benchmark configuration to create a queue benchmark:

```
POST http://localhost:5678/v1/bench
{
  "type": "queue",
  "operations": 10000,
  "concurrency": 10,
  "elementSize": 1024
}
```

Then the progress report and result are provided by the queue controller/executors as well:

```
GET http://localhost:5678/v1/bench/{benchId}/progress
```

The framework isn't currently dynamically extensible like other Atomix components are, but we can easily make it pluggable using the same patterns to allow custom benchmarks to be defined.

bench/src/main/java/io/atomix/bench/AtomixBenchmark.java
kuujo: Should probably be a warning against using benchmark nodes for anything other than benchmarking

kuujo:
I’m extending it a bit more. The end result will be that anyone can add a benchmark just by adding the relevant APIs, and any benchmark implementation can be run with arguments through the Python script.

I think we could also make an argument for putting this in a separate project, but I think it’s easiest to think about that once the code settles down.

kuujo:
@natraj09 maybe I will throw together a queue controller/executor tonight

natraj09:
Thanks @kuujo ! Really appreciate it.

kuujo:
So, there are three benchmark implementations currently:
* `log` tests a `DistributedLog` primitive with a `LogPartitionGroup`
* `map` tests an `AtomicMap` primitive with an arbitrary protocol
* `messaging` tests the Netty messaging layer

We could add one more primitive test for single partition/ordered primitives like `DistributedList` or `DistributedQueue`, but really what we're testing is the underlying protocols and partitioning schemes, so we should be able to produce the same numbers by just using e.g. a partition group configured with a single partition.

kuujo:
I think I'm actually going to move this to a separate project.

PR #955, opened 2018-12-14
Fix unstable log primitives test

kuujo:
Fixes a race condition in the distributed log primitives test. When the counter is created, if the counter receives log events prior to the value check then the value may not be `0`.


kuujo:
Actually, changing this patch into a larger change to how primitive sessions are established.

PR #974, opened 2019-01-16
Support singleton transactional primitives

kuujo:
This PR implements the correct behavior for `get()` when using transactional primitives. SIngleton transactional primitives are implemented by exposing a wrapped primitive to the `Transaction` to prevent it from closing the primitive sessions at the end of the transaction.


core/src/main/java/io/atomix/core/transaction/impl/DelegatingTransactionalMap.java
```diff
@@ -0,0 +1,87 @@
+package io.atomix.core.transaction.impl;
```
kuujo: Missing copyright info

PR #981, opened 2019-01-18
Initialize session with monotonically increasing service timestamp

kuujo:
This PR fixes a bug in how session timestamps are initialized. Currently, it's possible for a session to be immediately expired if a leader change occurs and the new leader has a clock that's behind the prior leader's last committed timestamp by more than the session timeout. This can result in the session being expired and then registered with the service.


kuujo:
This PR also introduces a delta that is computed when the service timestamp is decreased. The delta is used to compute the timestamp based on the rate at which the current leader's clock is progressing, but using the highest leader's clock as a reference.

PR #1005, opened 2019-03-05
Transition Raft servers to candidate/leader immediately if single node cluster

kuujo:
Just a small optimization for startup time in a single node cluster.


protocols/raft/src/main/java/io/atomix/protocols/raft/roles/CandidateRole.java
```diff
@@ -55,6 +55,11 @@ public CandidateRole(RaftContext context) {
 
   @Override
   public synchronized CompletableFuture<RaftRole> start() {
+    if (raft.getCluster().getActiveMemberStates().isEmpty()) {
+      log.debug("Single member cluster. Transitioning directly to leader.");
+      raft.transition(RaftServer.Role.LEADER);
```
menski: The new term and self vote still have to be persisted before transitioning to leader, i.e
```
raft.setTerm(raft.getTerm() + 1);
raft.setLastVotedFor(raft.getCluster().getMember().memberId());
```

Without this the raft is in an invalid state, that is also why the test are currently failing on this branch.

PR #1010, opened 2019-03-26
Use G1 GC in Atomix agent by default

kuujo:
We've seen significant improvements in stability in production using G1 rather than CMS. It pretty reliably keeps GC pauses below the threshold of default Raft leader election timeouts. This PR enables G1 with a 200ms pause goal.


johnou:
Have you considered Shenandoah GC? https://wiki.openjdk.java.net/display/shenandoah/Main

PR #1012, opened 2019-04-10
Accept raft role change listeners

Zelldon:


Zelldon:
Sorry this was wrong :see_no_evil:

Issue #1017, opened 2019-04-25
Raft leader enters an election loop when a follower restarts on kubernetes


deepthidevaki:
**Expected behavior**

Killing follower should not start an election. Even if it starts, the election should converge.

**Actual behavior**

When the pod of the follower is restarted, the leader steps down to a follower and restarts the election and then keeps transitioning between candidate -> leader -> follower -> candidate ..

Observation:
When we debugged it, we observed that when a pod is deleted, its dns address cannot be resolved sometime. Hence this [Address#address(boolean)](https://github.com/atomix/atomix/blob/master/utils/src/main/java/io/atomix/utils/net/Address.java#L167-L181) can return null. When  [messaging service](https://github.com/atomix/atomix/blob/master/cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyMessagingService.java#L486) resolves the address to null, it binds the address to the wildcard `0.0.0.0` which is the behavior of `InetSocketAddress`. So the messages that the leader sends to the dead node is received by the leader itself, which causes the leader to stepdown [here](https://github.com/atomix/atomix/blob/3a94b7c80576d762dd0d396d4645df07a0b37c31/protocols/raft/src/main/java/io/atomix/protocols/raft/roles/LeaderRole.java#L549).
There might be similar problem in [NettyUnicastService#unicast(Address, String, byte[])](https://github.com/atomix/atomix/blob/3a94b7c80576d762dd0d396d4645df07a0b37c31/cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyUnicastService.java#L85)

**Steps to reproduce**

Create a kubernet cluster using statefulset with replicas 3 and 
 Atomix configuration:
```
cluster {
        discovery {
            type: bootstrap
            nodes.1 {
              id: atomix-0
              address: "atomix-0.atomix.default.svc.cluster.local:5679"
            }
          }
    }
    managementGroup {
      type: raft
      partitions: 1
      storage.level: disk
      members: [atomix-1, atomix-2, atomix-0]
    }
    partitionGroups.consensus {
      type: raft
      partitions: 1
      partitionSize: 3
      storage.level: disk
      members: [atomix-1, atomix-2, atomix-0]
    }
```
Once all nodes have joined the Raft group, pick a node which is the follower for partition 1, for eg:- atomix-2. Delete pod for atomix-2 
`kubectl delete pod atomix-2`

**Environment**

 - Atomix: master (also observed in 3.1.5)
[...]

PR #1018, opened 2019-04-26
fix(cluster): do not send requests to unresolvable hosts

npepinpe:
Solves the issue mentioned in #1017, and prevents `NettyUnicastService` from mistakenly sending messages to itself when trying to send to an address which cannot be resolved. It's also possible there are other usages of `Address#address()` which are not handling the case where it's null.

### DIFF
--- a/cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyMessagingService.java
+++ b/cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyMessagingService.java
@@ -20,6 +20,7 @@ import javax.net.ssl.SSLException;
 import javax.net.ssl.TrustManagerFactory;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
+import java.net.InetAddress;
 import java.security.Key;
 import java.security.KeyStore;
 import java.security.MessageDigest;
@@ -470,6 +471,13 @@ public class NettyMessagingService implements ManagedMessagingService {
    */
   private CompletableFuture<Channel> bootstrapClient(Address address) {
     CompletableFuture<Channel> future = new OrderedFuture<>();
+    final InetAddress resolvedAddress = address.address(true);
+    if (resolvedAddress == null) {
+      future.completeExceptionally(new IllegalStateException("Failed to bootstrap client (address "
+          + address.toString() + " cannot be resolved)"));
+      return future;
+    }
+
     Bootstrap bootstrap = new Bootstrap();
     bootstrap.option(ChannelOption.ALLOCATOR, PooledByteBufAllocator.DEFAULT);
     bootstrap.option(ChannelOption.WRITE_BUFFER_WATER_MARK,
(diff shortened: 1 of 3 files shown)


cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyUnicastService.java
```diff
@@ -78,11 +80,17 @@ public NettyUnicastService(Address address, MessagingConfig config) {
...
+    if (resolvedAddress == null) {
+      LOGGER.debug("Cannot send unicast message to destination address {} as it cannot be resolved", address);
+      return;
+    }
+
     Message message = new Message(this.address, subject, payload);
     byte[] bytes = SERIALIZER.encode(message);
     ByteBuf buf = channel.alloc().buffer(4 + bytes.length);
     buf.writeInt(bytes.length).writeBytes(bytes);
-    channel.writeAndFlush(new DatagramPacket(buf, new InetSocketAddress(address.address(), address.port())));
```
johnou: looks like this may fix a memory leak if writeAndFlush does not consume the bytebuf, nice!

johnou:
@npepinpe please rebase so CI passes.

menski:
@johnou I rebased the branch, as @npepinpe is on vacation. Let me know if we should do anything else before merge. thanks

cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyMessagingService.java
```diff
@@ -470,6 +471,14 @@ public void unregisterHandler(String type) {
    */
   private CompletableFuture<Channel> bootstrapClient(Address address) {
     CompletableFuture<Channel> future = new OrderedFuture<>();
+    final InetAddress resolvedAddress = address.address(true);
+    if (resolvedAddress == null) {
+      final String errorMessage = String.format("Cannot bootstrap client for address %s as it cannot be resolved",
+              address.toString());
+      future.completeExceptionally(new IllegalStateException(errorMessage));
```
johnou: can we simplify this to something like

```java
future.completeExceptionally(new IllegalStateException("Failed to bootstrap client (address " 
                                                             + address.toString() + " cannot be resolved)"));
```

cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyUnicastService.java
```diff
@@ -78,11 +80,17 @@ public NettyUnicastService(Address address, MessagingConfig config) {
 
   @Override
   public void unicast(Address address, String subject, byte[] payload) {
+    final InetAddress resolvedAddress = address.address();
+    if (resolvedAddress == null) {
+      LOGGER.debug("Cannot send unicast message to destination address {} as it cannot be resolved", address);
```
johnou:
```
"Failed sending unicast message (destination address {} cannot be resolved)"
```

menski:
@johnou I updated the messages

johnou:
Thanks again!

saravanandls:
I am facing similar issue where leader election goes to loop when a follower is down. From which version this fix is available. I am using atomix - 3.0.11 docker image.

PR #1027, opened 2019-06-04
Reset member snapshot index if outdated

npepinpe:
Fixes an issue where if a follower suffers data loss before a new snapshot is taken on the leader, the snapshot is not resent, as the leader "remembers" the snapshot index it previously installed on the follower. Also a slight optimization in the general case of data loss on follower nodes since their snapshot index is automatically updated.

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/roles/AbstractAppender.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/roles/AbstractAppender.java
@@ -238,6 +238,7 @@ abstract class AbstractAppender implements AutoCloseable {
     else {
       resetMatchIndex(member, response);
       resetNextIndex(member, response);
+      resetSnapshotIndex(member, response);
 
       // If there are more entries to send then attempt to send another commit.
       if (response.lastLogIndex() != request.prevLogIndex() && hasMoreEntries(member)) {
@@ -311,6 +312,17 @@ abstract class AbstractAppender implements AutoCloseable {
     if (member.getLogReader().getNextIndex() != nextIndex) {
       member.getLogReader().reset(nextIndex);
       log.trace("Reset next index for {} to {}", member, nextIndex);
+    }
+  }
+
+  /**
+   * Resets the snapshot index of the member when a response fails.
+   */
+  protected void resetSnapshotIndex(RaftMemberContext member, AppendResponse response) {
+    final long snapshotIndex = response.lastSnapshotIndex();
+    if (member.getSnapshotIndex() != snapshotIndex) {
+      member.setSnapshotIndex(snapshotIndex);
+      log.trace("Reset snapshot index for {} to {}", member, snapshotIndex);
     }
   }
 
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/roles/AbstractAppender.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/roles/AbstractAppender.java
@@ -322,7 +322,7 @@ abstract class AbstractAppender implements AutoCloseable {
     final long snapshotIndex = response.lastSnapshotIndex();
     if (member.getSnapshotIndex() != snapshotIndex) {
       member.setSnapshotIndex(snapshotIndex);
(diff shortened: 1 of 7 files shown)


menski:
done :+1:

protocols/raft/src/test/java/io/atomix/protocols/raft/RaftTest.java
```diff
@@ -1372,6 +1487,13 @@ private TestPrimitive createPrimitive(RaftClient client, ReadConsistency consist
     return new TestPrimitiveImpl(proxy, registry);
   }
 
+  private CompletableFuture<Void> reduceFutures(Stream<CompletableFuture> stream) {
```
johnou: couldn't we use something like this?

```
    public static <F extends CompletableFuture<?>> CompletableFuture<Void> allOf(Stream<F> cfs) {
        return CompletableFuture.allOf(cfs.toArray(CompletableFuture[]::new));
    }
```

menski: I changed it to use instead https://github.com/atomix/atomix/blob/dbad13756abfd997386c2d046a3d8bb8d45570d3/utils/src/main/java/io/atomix/utils/concurrent/Futures.java#L175

johnou:
I'll defer the review to @kuujo for this one, he knows the Raft system best.

kuujo:
I think the issue and the solution make sense. If the leader is designed to be able to resend entries rather than considering the follower corrupt, it _must_ also resend snapshots.

But I’m wondering if another possible solution is to just use the follower’s `lastLogIndex` to reset the leader’s `snapshotIndex` for the follower if `lastLogIndex < snapshotIndex`. Snapshots are essentially sent when they’re encountered at some index during log replication, so it may be reasonable to just assume if a follower’s `lastLogIndex` is less than any given snapshot index, the follower doesn’t have that snapshot and it needs to be sent. That should cover the case where logs and snapshots are lost, though not when just snapshots are lost.

BTW we should probably be printing a stern warning when the leader detects data missing on a follower.

deepthidevaki:
> But I’m wondering if another possible solution is to just use the follower’s `lastLogIndex` to reset the leader’s `snapshotIndex` for the follower if `lastLogIndex < snapshotIndex`.

We had a similar solution at first where we reset `snapshotIndex` to 0 (resetting to `lastLogIndex` would have been better). If the follower is just too slow or the network is poor that it drops many append requests, there will be many failed `AppendResponse`. In this case there is no data loss, but the leader resets the indexes for every failed response. Then, I observed that the same snapshot is resent multiple times, because the leader unnecessarily resets the `snapshotIndex`. I think it won't lead to any inconsistent state. But it will unnecessarily flood the already slow follower/network. That is why @npepinpe went for the current solution. 

If my understanding of `LeaderAppender#appendEntries` code is correct, this can also happen when we reset `snapshotIndex` to `lastLogIndex`.  Because a snapshot can be sent well before the log replication reaches the `snapshotIndex`. (`if (snapshot != null && member.getSnapshotIndex() < snapshot.index() && snapshot.index() >= member.getLogReader().getCurrentIndex())`)

protocols/raft/src/main/java/io/atomix/protocols/raft/roles/AbstractAppender.java
```diff
@@ -314,6 +315,17 @@ protected void resetNextIndex(RaftMemberContext member, AppendResponse response)
...
   }
 
+  /**
+   * Resets the snapshot index of the member when a response fails.
+   */
+  protected void resetSnapshotIndex(RaftMemberContext member, AppendResponse response) {
+    final long snapshotIndex = response.lastSnapshotIndex();
+    if (member.getSnapshotIndex() != snapshotIndex) {
+      member.setSnapshotIndex(snapshotIndex);
+      log.trace("Reset snapshot index for {} to {}", member, snapshotIndex);
```
johnou: could be warn as @kuujo suggested?

johnou:
lgtm @kuujo are further changes required or can we pull this in?

johnou:
@kuujo ptal

johnou:
https://github.com/atomix/atomix/pull/1069

PR #1031, opened 2019-06-08
SonarQube - String Literals on the left on comparisons

dvmarcilio:
Transforming from `someString.equals("string")`  to `"string".equals(someString)`. This avoids checking for `null` and also prevents `NullPointerException` being thrown.

This PR fixes SonarQube violations of the rule: [Strings literals should be placed on the left side when checking for equality](https://rules.sonarsource.com/java/RSPEC-1132)


johnou:
@dvmarcilio thanks!

PR #1032, opened 2019-06-08
SonarQube - Replacing String comparison from '!=' to '!.equals()'

dvmarcilio:
SonarQube rule states that is almost always a mistake to compare two instances of `String` with `==`, because it compares location in memory. `equals()` should be used.

Here is the link for the rule: [Strings and Boxed types should be compared using "equals()"](https://rules.sonarsource.com/java/RSPEC-4973)


dvmarcilio:
Done!

johnou:
Thanks!

PR #1040, opened 2019-07-02
Don't commit non replicated entries in old leader upon reconnecting

### LINKED ISSUE #1036: Uncommitted events can become committed without quorum
**Expected behavior**

When an isolated node rejoins a raft, it should not commit any entries that were not replicated to other nodes.

**Actual behavior**

When a node that is leader for a partition becomes isolated (i.e. network partition), it can have written events to its log that were not replicated and therefore not committed (but still present, as it will keep trying to append them to its unreachable followers).

When the other nodes keep replicating amongst themselves and compaction occurs, and the isolated node rejoins, it will receive an empty append request from the new leader (i.e. a heartbeat) which contains the new commit index, term, previous log index, and previous log term. The new commit index is definitely higher than the index of the uncommitted leftover entries, and if truncation occurred on the new leader, it's possible that the previous log term is 0.

When the append request has no entries to append, and the previous log term is 0, then the append request is not rejected and the follower blindly commits all entries up to the new commit index, therefore committing the leftover entries.

**Steps to reproduce**

It's a bit hard, but we could reliably reproduce it by stress testing our system and disconnecting one node from the others, then ensure one or more backups is taken on the remaining nodes before reconnecting the isolated node; leftover entries will be applied to the primitives. This is particularly noticeable for us since the state of our custom primitive is stored on disk.

None yet, will try to come up with a test

**Environment**

 - Atomix: 3.2.0
 - OS: Linux
 - JVM: java version "1.8.0_181" HotSpot

deepthidevaki:
This PR fixes #1036 
 * added a test to verify that non replicated events can get committed when old leader reconnects after a network partition.

To fix the issue, following changes are made to the raft implementation:
 * previous term is 0 only for the first entry in the log
 * if there is no previous entry in the log after compaction, use the existing snapshot as previous entry.
 * reset follower's next index when install response is success.
 * reset log after receiving a snapshot from leader.
 * include the last applied entry's term in the snapshot
 * replaces SnapshotDescriptor positions and lengths with constants
       similar to JournalSegmentDescriptor

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/SnapshotDescriptor.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/SnapshotDescriptor.java
@@ -30,7 +30,23 @@ import static com.google.common.base.Preconditions.checkNotNull;
  */
 public final class SnapshotDescriptor implements AutoCloseable {
   public static final int BYTES = 64;
-  public static final int VERSION = 1;
+
+  // Current snapshot version.
+  private static final int VERSION = 1;
+
+  // The lengths of each field in the header.
+  private static final int INDEX_LENGTH = Long.BYTES;          // 64-bit signed integer
+  private static final int TIMESTAMP_LENGTH = Long.BYTES;      // 64-bit signed integer
+  private static final int VERSION_LENGTH = Integer.BYTES;     // 32-bit signed integer
+  private static final int LOCKED_LENGTH = 1;                  // 8-bit signed byte
+  private static final int TERM_LENGTH = Long.BYTES;           // 64-bit signed integer
+
+  // The positions of each field in the header.
+  private static final int INDEX_POSITION = 0;                                           // 0
+  private static final int TIMESTAMP_POSITION = INDEX_POSITION + INDEX_LENGTH;           // 8
+  private static final int VERSION_POSITION = TIMESTAMP_POSITION + TIMESTAMP_LENGTH;     // 16
+  private static final int LOCKED_POSITION = VERSION_POSITION + VERSION_LENGTH;          // 20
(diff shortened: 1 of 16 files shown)


johnou:
@deepthidevaki would you kindly rebase?

deepthidevaki:
rebased. :+1:

protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/SnapshotDescriptor.java
```diff
@@ -30,7 +30,23 @@
  */
 public final class SnapshotDescriptor implements AutoCloseable {
   public static final int BYTES = 64;
-  public static final int VERSION = 1;
+
+  // Current snapshot version.
+  private static final int VERSION = 1;
```
johnou: one observation here is that version was never incremented and there is no backwards compatibility with version one, are snapshots always storage level memory? maybe it's not an actual issue.

johnou:
@kuujo ptal

protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/SnapshotDescriptor.java
johnou: @deepthidevaki how has this been performing in production, did you see this comment?

johnou:
https://github.com/atomix/atomix/pull/1070

protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/SnapshotDescriptor.java
deepthidevaki: Sorry. Yes. Compatibility could be an issue. We never ran into that because we never had to upgrade from previous version.

PR #1043, opened 2019-07-11
fix(raft): fixes issues with concurrent snapshot writers

### LINKED ISSUE #1042: Concurrent writes to Raft snapshot files creates unusable snapshots
**Expected behavior**

Snapshots replicated from the leader should be an exact copy of the leader's snapshot, and locally written snapshots should be readable.

**Actual behavior**

It is possible that we schedule compaction for index X (on the Raft state thread) while at the same time replicating the snapshot for index X from our leader (on the Raft thread context). Since both point to the same file, and the writer lock is on the `FileSnapshot` instance, they can both write to the same file. This is exacerbated by the fact that on the state thread, when taking a snapshot, we schedule completion for later, so locking the snapshot is asynchronous. 

**Steps to reproduce**

Our setup (though may be reproduceable with smaller setups):

- Cluster of 5 nodes
- 8 partitions
- Replication factor of 3

Add logging when before/after persisting an in memory snapshot, and logging as well when completing the snapshot in `RaftServiceManager#scheduleCompletion` right before `snapshot.complete()` is called.

Add logging in `PassiveRole` before/after `pendingSnapshot.commit()` in `PassiveRole#onInstall`.

Start cluster, run some load, and randomly kill the leaders to cause fail over. If the planets align, you will sometimes see the logging statements above interleaved showing that we access the same file from two different threads.

Symptoms caused by this are unreadable snapshots (e.g. [this issue](https://github.com/atomix/atomix/issues/1007) or [this](https://github.com/zeebe-io/zeebe/issues/2766), and [this](https://github.com/zeebe-io/zeebe/issues/2359)).

None but if need be I can try to come up with one.

**Environment**

 - Atomix: 3.2.0 (master)
 - OS: Linux
 - JVM: Java 1.8.0_181

npepinpe:
- removes the concept of "persisting" a snapshot; with in memory storage level, snapshots are not persisted, and with anything else they are written to a temporary file before being moved to a permanent location on complete
- file snapshots read from the permanent file (or fail if none yet present)
- closing a file snapshot discards its temporary file
- improves error handling on snapshot completion failure (when replicating and locally snapshotting)

Closes #1042

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/FileSnapshot.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/FileSnapshot.java
@@ -17,10 +17,12 @@ package io.atomix.protocols.raft.storage.snapshot;
 
 import io.atomix.storage.buffer.Buffer;
 import io.atomix.storage.buffer.FileBuffer;
+import io.atomix.utils.AtomixIOException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 
@@ -42,7 +44,9 @@ final class FileSnapshot extends Snapshot {
   @Override
   public synchronized SnapshotWriter openWriter() {
     checkWriter();
-    Buffer buffer = FileBuffer.allocate(file.file(), SnapshotDescriptor.BYTES);
+    checkState(!file.file().exists(), "cannot write to completed snapshot");
+    checkNotNull(file.temporaryFile(), "missing temporary snapshot file for writing");
+    Buffer buffer = FileBuffer.allocate(file.temporaryFile(), SnapshotDescriptor.BYTES);
     descriptor.copyTo(buffer);
 
     int length = buffer.position(SnapshotDescriptor.BYTES).readInt();
@@ -66,17 +70,36 @@ final class FileSnapshot extends Snapshot {
   }
 
   @Override
-  public boolean isPersisted() {
-    return true;
-  }
-
-  @Override
   public Snapshot complete() {
-    Buffer buffer = FileBuffer.allocate(file.file(), SnapshotDescriptor.BYTES);
+    checkNotNull(file.temporaryFile(), "no temporary snapshot file to read from");
+
+    Buffer buffer = FileBuffer.allocate(file.temporaryFile(), SnapshotDescriptor.BYTES);
     try (SnapshotDescriptor descriptor = new SnapshotDescriptor(buffer)) {
       descriptor.lock();
     }
+
+    try {
+      Files.move(file.temporaryFile().toPath(), file.file().toPath());
+    } catch (FileAlreadyExistsException e) {
+      LOGGER.debug("Snapshot {} was already previously completed", this);
+    } catch (IOException e) {
+      throw new AtomixIOException(e);
+    }
+
+    file.clearTemporaryFile();
     return super.complete();
+  }
+
+  /**
+   * Deletes the temporary file
+   */
+  @Override
+  public void close() {
+    super.close();
+
+    if (file.temporaryFile() != null) {
+      deleteFileSilently(file.temporaryFile().toPath());
+    }
   }
 
   /**
@@ -86,11 +109,17 @@ final class FileSnapshot extends Snapshot {
   public void delete() {
     LOGGER.debug("Deleting {}", this);
     Path path = file.file().toPath();
+
     if (Files.exists(path)) {
-      try {
-        Files.delete(file.file().toPath());
-      } catch (IOException e) {
-      }
+      deleteFileSilently(path);
+    }
+  }
+
+  private void deleteFileSilently(Path path) {
+    try {
+      Files.delete(path);
+    } catch (IOException e) {
+      LOGGER.debug("Failed to delete snapshot file {}", path, e);
     }
   }
 }
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/FileSnapshot.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/FileSnapshot.java
@@ -25,6 +25,8 @@ import java.io.IOException;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.nio.file.AtomicMoveNotSupportedException;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
@@ -79,9 +81,7 @@ final class FileSnapshot extends Snapshot {
     }
 
     try {
-      Files.move(file.temporaryFile().toPath(), file.file().toPath());
-    } catch (FileAlreadyExistsException e) {
-      LOGGER.debug("Snapshot {} was already previously completed", this);
+      persistTemporaryFile();
     } catch (IOException e) {
       throw new AtomixIOException(e);
     }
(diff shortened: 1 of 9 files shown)


protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/SnapshotFile.java
```diff
@@ -90,6 +93,19 @@ static File createSnapshotFile(File directory, String serverName, long index) {
     return new File(directory, createSnapshotFileName(serverName, index));
   }
 
+  /**
+   * Creates a temporary file for writing snapshots.
+   */
+  static File createTemporaryFile(File base) {
+    try {
+      final File file = File.createTempFile(base.getName(), null);
+      file.deleteOnExit();
```
johnou: @npepinpe please replace this method call with http://docs.oracle.com/javase/7/docs/api/java/nio/file/StandardOpenOption.html#DELETE_ON_CLOSE to avoid possible memory leak with shutdown hooks.

johnou: Or better yet can we just delete it when done with it? There's a few SO threads where they say the file is deleted immediately on some systems.

protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/FileSnapshot.java
```diff
@@ -65,32 +69,57 @@ public synchronized SnapshotReader openReader() {
...
-    Buffer buffer = FileBuffer.allocate(file.file(), SnapshotDescriptor.BYTES);
+    checkNotNull(file.temporaryFile(), "no temporary snapshot file to read from");
+
+    Buffer buffer = FileBuffer.allocate(file.temporaryFile(), SnapshotDescriptor.BYTES);
     try (SnapshotDescriptor descriptor = new SnapshotDescriptor(buffer)) {
       descriptor.lock();
     }
+
+    try {
+      Files.move(file.temporaryFile().toPath(), file.file().toPath());
```
johnou: Can we first try with StandardCopyOption.ATOMIC_MOVE then fallback with no option?

johnou: Assuming it works with TMP files.

npepinpe: I think the main issue here is that `ATOMIC_MOVE` doesn't work well across file systems where it needs to actually be copied; I think setting up your system temp directory as something like `ramfs` or `tmpfs` is fairly common, so I would expect it to fail. We can keep the current solution as fallback, but I guess I would expect the fallback to be used more often; I have no proof for this though.

protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/SnapshotFile.java
npepinpe: The `deleteOnExit` here is just a failsafe in case the snapshot isn't closed; the temp file is removed when closing the snapshot or completing it. `DELETE_ON_CLOSE` seems more robust than `File.deleteOnExit()` though, so I'll update it to that.

> Or better yet can we just delete it when done with it? There's a few SO threads where they say the file is deleted immediately on some systems.

Do you mean that opening it with `DELETE_ON_CLOSE` causes it to delete the file prematurely?

protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/FileSnapshot.java
johnou: Yeah that's where I was going with my previous comment but I'm traveling right now so difficult to verify.

protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/SnapshotFile.java
johnou: Might be a little easier for you to verify but is the shutdown hook removed if the file is closed?

protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/FileSnapshot.java
npepinpe: I verified on my own system where `/tmp` is mounted as a `tmpfs` partition and the Raft log is on a different standard `ext4` partition. What I meant by no proof is I don't really know that most production setups will have `tmp` on a different partition with a different file system, but if I had to guess I'd say most likely.

protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/SnapshotFile.java
npepinpe: Good catch, it's definitely not removed. `DELETE_ON_CLOSE` would be preferable (though I'm not sure how it's implemented). But I think using `DELETE_ON_CLOSE` means refactoring `FileBytes` to use NIO channels instead of `RandomAccessFile`. In either case I think `deleteOnExit` is not really all that useful then.

protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/FileSnapshot.java
johnou: I suppose we could pass in the third argument directory and use the storage path along with StandardCopyOption.ATOMIC_MOVE and fallback to cut down on IO?

protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/SnapshotFile.java
johnou: Might be easier to just make a best effort that the file is deleted when we are done with it then.

npepinpe: The file is deleted when the snapshot is closed or completed. Snapshots loaded from disks do not have a temp file associated, so they also leave nothing behind. `deleteOnClose` in this case is only useful if we introduce an error where we forget to close or complete snapshots, since it doesn't work if the JVM crashes. I could take it or leave it (leaning more towards leaving it imo), up to you.

protocols/raft/src/main/java/io/atomix/protocols/raft/storage/snapshot/FileSnapshot.java
npepinpe: It might be better to use the storage path anyway, as large snapshots could eat up quite a lot of RAM if your temp directory is mounted using `tmpfs`. I guess we can use UUIDs to keep the path unique and avoid two instances from two different threads writing to the same file (since afaik file locks are only helpful when dealing with multiple JVMs). Or do you have a another idea?

johnou: sgtm, but I am not sure if the uuid is necessary? the first arg is prefix, Java adds some extra random chars with createTemp.. https://docs.oracle.com/javase/7/docs/api/java/io/File.html#createTempFile(java.lang.String,%20java.lang.String,%20java.io.File)

npepinpe: Yeah you're right, seems pretty unlikely you'd get collisions.

johnou:
@npepinpe looks like CI unit test failed.. related?

ERROR] Failures: 
[ERROR]   RaftTest.lambda$testSessionDelete$30:1201->ConcurrentTestCase.threadAssertTrue:48 expected true
[INFO] 
[ERROR] Tests run: 107, Failures: 1, Errors: 0, Skipped: 1

johnou:
Build passed on retry.

johnou:
@kuujo ptal

johnou:
@npepinpe how has this been performing in production? any follow up patches necessary?

johnou:
@kuujo ptal

johnou:
@npepinpe I take it this has been working well for you guys in production?

johnou:
I took a second pass, looks good to me.

johnou:
@npepinpe thanks a lot!

PR #1045, opened 2019-07-17
fix(raft): fixes race condition with setRequestSequence

### LINKED ISSUE #1044: Race condition in RaftSession#setRequestSequence
**Expected behavior**

The request sequence of a session is always increasing.

**Actual behavior**

Both the raft thread context and the state thread context will access the same session instance and call `setRequestSequence`, which is a non-volatile field, and perform a non-atomic operation `this.requestSequence = Math.max(this.requestSequence, requestSequence)`. 

The `LeaderRole` will update the session's `requestSequence` in the raft thread context (e.g. `onCommand`, `drainCommands`, etc.) and the `RaftServiceContext` will update it on the state thread context. As the `RaftServiceContext` is usually behind in applying entries, then it's possible that, with unlucky scheduling, calling `registerResult` will update the session's `requestSequence` to a lower sequence.

This can result in requests appearing as "missing" in `LeaderRole`, for example in `LeaderRole#onCommand`, where the session's `nextRequestSequence` is compared with the current command's `sequenceNumber`. 

**Steps to reproduce**

As it is a scheduling issue, it's kind of hard to reproduce. To reproduce it I tracked in parallel every session's `requestSequence` in `LeaderRole` and also tracked every processed request, and when a request was deemed "missing" (e.g. in `LeaderRole.java:629`) checked if the request has previously been seen (and if so rejected or not) and compared what the current sequence number of the session was with the one I was tracking manually. All this was tested on a single node, single partition setup, so it was using a local connection and could not have "lost" requests.

As it's a race condition it's very hard to write a reliable reproducer for it. I could reproduce it I'd say ~30% of the time using the following [gist](https://gist.github.com/npepinpe/77f7bc6fc9e3ed23d879d8b37465ded9); if you debug it and break in `LeaderRole.java` right before rejecting the request for having too high of a sequence number (around line 629), i.e. in:

```java
if (sequenceNumber > session.nextRequestSequence()) {
   // break here or add debug logging
}
```

You should see this happen at times, even though we're in a single node/partition set-up using local connections.

**Environment**

 - Atomix: 3.2.0
 - OS: Linux 5.1.16
 - JVM: java version "1.8.0_181"

npepinpe:
- `RaftSession#requestSequence` is now a `LongAccumulator` which tracks the max of request sequences

Closes #1044

### DIFF
--- a/protocols/raft/src/main/java/io/atomix/protocols/raft/session/RaftSession.java
+++ b/protocols/raft/src/main/java/io/atomix/protocols/raft/session/RaftSession.java
@@ -45,6 +45,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Queue;
+import java.util.concurrent.atomic.LongAccumulator;
 
 import static com.google.common.base.MoreObjects.toStringHelper;
 import static com.google.common.base.Preconditions.checkState;
@@ -63,7 +64,7 @@ public class RaftSession extends AbstractSession {
   private final ThreadContext eventExecutor;
   private volatile State state = State.CLOSED;
   private volatile long lastUpdated;
-  private long requestSequence;
+  private LongAccumulator requestSequence;
   private volatile long commandSequence;
   private volatile long lastApplied;
   private volatile long commandLowWaterMark;
@@ -106,6 +107,7 @@ public class RaftSession extends AbstractSession {
         .add("type", context.serviceType())
         .add("name", context.serviceName())
         .build());
+    this.requestSequence = new LongAccumulator(Long::max, 0);
   }
 
   /**
@@ -196,7 +198,7 @@ public class RaftSession extends AbstractSession {
    * @return The session request number.
    */
   public long getRequestSequence() {
-    return requestSequence;
+    return requestSequence.get();
   }
 
   /**
@@ -205,7 +207,7 @@ public class RaftSession extends AbstractSession {
    * @return the next request sequence number
    */
   public long nextRequestSequence() {
-    return this.requestSequence + 1;
+    return this.requestSequence.get() + 1;
   }
 
   /**
@@ -214,7 +216,7 @@ public class RaftSession extends AbstractSession {
    * @param requestSequence the current request sequence number
    */
   public void setRequestSequence(long requestSequence) {
-    this.requestSequence = Math.max(this.requestSequence, requestSequence);
+    this.requestSequence.accumulate(requestSequence);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


npepinpe:
Two things:

1. I notice `RaftSession#setRequestSequence` and `RaftSession#resetRequestSequence` are essentially the same (and also used to be); I kept them as I imagine they were added for semantics, but they could just be merged.
1. I didn't add a test as I wasn't sure how to write a reliable one, since the race conditions relies on both threads thinking they're setting the "max" value, so would require (?) too fine grained scheduling I think. If you have any ideas I'll gladly add one!

npepinpe:
Build is not passing but seems to be related to #1041

protocols/raft/src/main/java/io/atomix/protocols/raft/session/RaftSession.java
```diff
@@ -63,7 +64,7 @@
   private final ThreadContext eventExecutor;
   private volatile State state = State.CLOSED;
   private volatile long lastUpdated;
-  private long requestSequence;
+  private LongAccumulator requestSequence;
```
johnou: Can this be final?

protocols/raft/src/main/java/io/atomix/protocols/raft/session/RaftSession.java
```diff
@@ -205,7 +207,7 @@ public long getRequestSequence() {
    * @return the next request sequence number
    */
   public long nextRequestSequence() {
-    return this.requestSequence + 1;
+    return this.requestSequence.get() + 1;
```
johnou: get isn't thread safe according to the javadocs, https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/LongAccumulator.html#get--  then again the previous implementation wasn't thread safe either.

npepinpe: Hm, I originally figured that it's fine if it doesn't return the latest value, and more importantly just that it cannot set a lower value. I think in most cases it should be fine, but that probably breaks the principle of least astonishment. Easy fix here would be to convert to an `AtomicLong`.

johnou: Or use a volatile long with AtomicLongFieldUpdater?

npepinpe: I'm not really familiar with using `AtomicLongFieldUpdater`; what's the difference with `AtomicLong`? Isn't an `AtomicLong` just a volatile long with some CAS methods?

protocols/raft/src/main/java/io/atomix/protocols/raft/session/RaftSession.java
npepinpe: Yes of course, my bad 👍

johnou:
Thanks!

PR #1054, opened 2019-10-10
fix(protocols/raft): apply raft partition group configuration

saig0:
Apply raft configuration instead of using hardcoded values for election timeout, heartbeat interval and session timeout.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ services:
   - docker
 
 jdk:
-- oraclejdk8
 - openjdk8
 - openjdk10
 - openjdk11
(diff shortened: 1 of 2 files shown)


johnou:
@saig0 are the defaults in the builder the same?

saig0:
Yes, the defaults are the same.

johnou:
@saig0 thanks!

marcinostry:
hello @johnou @saig0 
Do you know when tagged version contains the fix will be released? I can see the latest tagged version is 3.1.5. Should I expect 3.1.6 in near future?

PR #1076, opened 2020-06-24
Bulk update mmap

pierventre:


core/src/main/java/io/atomix/core/multimap/impl/AbstractAtomicMultimapService.java
```diff
@@ -252,6 +252,23 @@ public boolean removeAll(String key, Collection<? extends byte[]> values) {
...
+  public boolean removeAll(Map<String, Collection<? extends byte[]>> mapping) {
+    // There are no updates
+    if (mapping.isEmpty()) {
+      return false;
+    }
+
+    // Decompose the commit in several updates
+    boolean operationResult = false;
+    for (Map.Entry<String, Collection<? extends byte[]>> entry : mapping.entrySet()) {
+      if(removeAll(entry.getKey(), entry.getValue())) {
```
kuujo, line 265: I think this code needs to be formatted. Should be a space after the `if`

kuujo (changes requested):

kuujo (approved):
I think you got it!

Issue #1058, opened 2019-12-10
VERSION classpath resource conflicts with other libraries


benji:
The version is read from `Atomix.VERSION_RESOURCE = "VERSION"` which may conflict with other libs like org.openjsse:openjsse:1.1.0.

We should either rename it to `ATOMIX_VERSION` or move it to an atomix package like `io/atomix/VERSION`.

ynfeng:
I have sae issue.

PR #1080, opened 2020-08-11
Fix VERSION resource conflicts.

ynfeng:
Ref to #1058


johnou:
Personally I think it's nicer if we move it to a package instead eg. io/atomix/VERSION

johnou:
@ynfeng please move the file to the new path as per my review.

ynfeng:
> @ynfeng please move the file to the new path as per my review.

ok

ynfeng:
@johnou done

ynfeng:
@johnou done.

johnou (approved):

PR #1081, opened 2020-08-13
Fix for TLS configuration

pierventre:
The `logKeyStore` function was picking in a wrong way the keys.
TrustedCertEntry cannot be retrieved through `getKey` function.

### DIFF
--- a/cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyMessagingService.java
+++ b/cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyMessagingService.java
@@ -200,15 +200,17 @@ public class NettyMessagingService implements ManagedMessagingService {
       try {
         for (Enumeration<String> e = ks.aliases(); e.hasMoreElements(); ) {
           String alias = e.nextElement();
-          Key key = ks.getKey(alias, ksPwd);
+          // Use this for PrivateKeyEntry and SecretKeyEntry
           Certificate[] certs = ks.getCertificateChain(alias);
           log.debug("{} -> {}", alias, certs);
           final byte[] encodedKey;
+          // If the key is imported this check will fail
           if (certs != null && certs.length > 0) {
             encodedKey = certs[0].getEncoded();
           } else {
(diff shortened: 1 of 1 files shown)


kuujo (approved):

cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyMessagingService.java
```diff
@@ -200,15 +200,17 @@ private void logKeyStore(KeyStore ks, String ksLocation, char[] ksPwd) {
       try {
         for (Enumeration<String> e = ks.aliases(); e.hasMoreElements(); ) {
           String alias = e.nextElement();
-          Key key = ks.getKey(alias, ksPwd);
```
johnou, line 203: this change will break keystores with passwords.

johnou: please add unit test with keystore that contains password protected cert that expires in 100~ years

pierventre: > this change will break keystores with passwords.

Are you sure of this ? I believe we tried with password protect keystores. Do you have any example of unit tests ?

johnou: I'll grab an example on Monday when I'm connected to work VPN, recently integrated password protected certificates in one of our apps.

pierventre: @johnou thanks a lot!

pierventre: @johnou ping

johnou: @pierventre apologies for the satellite delay.. https://gist.github.com/johnou/d9118891ada6791717ea322242cd497a

PR #1087, opened 2020-09-29
Avoid unnecessary transfer and notify behavior when execute rebalance

hantangwangd:
Find and transfer leadership to the candidate with the fewest primaries directly, avoid unnecessary transfer and notify behavior


storage/src/test/java/io/atomix/storage/buffer/BufferTest.java
```diff
@@ -374,6 +376,30 @@ public void testCompact() {
     assertEquals(1234, buffer.readLong());
   }
 
+  @Test
+  public void testCompactWithOffsetNotZero() {
```
johnou, line 380: nice catch cc @kuujo

hantangwangd:
what do you think? @johnou @jhalterman

PR #1091, opened 2021-01-23
Fix for TLS configuration

pierventre:
The `logKeyStore` function was picking in a wrong way the keys. TrustedCertEntry cannot be retrieved through `getKey` function. Includes unit tests for password protected keystores that contains password protected certs expiring in 100 years.


pierventre:
@johnou, let me know if the tests are ok

PR #1095, opened 2021-04-14
Leak on server shutdown while still awaiting other nodes to join

### LINKED ISSUE #1096: Leak on server shutdown while still awaiting other nodes to join
**Expected behavior**

Restarting `Atomix` configured to use N raft nodes, bit still not connected to any nodes, should not leak any `Atomix` instance.

**Actual behavior**

`DefaultRaftServer::shutdown()` is not closing it's `RaftContext` and cleaning up `RaftContext::threadContext` tasks.

**Steps to reproduce**

Create a single raft node (expecting N nodes in total), start it and stop it right after, while awaiting it to stop.
After running a full GC, the `Atomix` instance is leaking (some screenshots below).

```java
   public static Atomix createAtomix(String localMemberId,
                                     File dataDirectory,
                                     Map<String, Address> nodes) {
      final Address localAddress = nodes.get(localMemberId);
      if (localAddress == null) {
         throw new IllegalArgumentException("the local member id should been included in the node map");
      }
      final AtomixBuilder atomixBuilder = Atomix.builder().withMemberId(localMemberId).withAddress(localAddress);
      atomixBuilder
         .withMembershipProvider(BootstrapDiscoveryProvider.builder()
                                    .withNodes(
                                       nodes.entrySet().stream()
                                          .map(entry-> Node.builder()
                                             .withId(entry.getKey())
                                             .withAddress(entry.getValue())
                                             .build())
                                          .collect(Collectors.toList())).build());
      // using Profile.consensus(members) is a short-cut of this but it won't left any config choice
      atomixBuilder
         .withManagementGroup(
            RaftPartitionGroup.builder("system")
               .withNumPartitions(1)
               .withMembers(nodes.keySet())
               .withStorageLevel(StorageLevel.DISK)
               .withDataDirectory(new File(dataDirectory, "management"))
               .build())
         .withPartitionGroups(
            RaftPartitionGroup.builder("data")
               .withNumPartitions(1)
               .withMembers(nodes.keySet())
               .withStorageLevel(StorageLevel.DISK)
               .withDataDirectory(new File(dataDirectory, "data"))
               .build());
      return atomixBuilder.build();
   }

   @Test
   public void atomixLeak() {
      File f = new File("./atomix");
      f.deleteOnExit();
      final String localId = "a";
      final Address localAddress = Address.from("localhost:7070");
      final Map<String, Address> nodes = new HashMap<>(3);
      nodes.put(localId, localAddress);
      nodes.put("b", Address.from("localhost:7071"));
      nodes.put("c", Address.from("localhost:7072"));
      Atomix atomix = createAtomix("a", f, nodes);
      try {
         // wait a bit in order to get the RaftServer::start called
         atomix.start().get(2, TimeUnit.SECONDS);
         Assert.fail();
      } catch (TimeoutException te) {
         try {
            atomix.stop().join();
         } catch (Throwable t) {
            Assert.fail();
         }
      } catch (Throwable t) {
         Assert.fail();
      }
   }
```
It's important to take an heap snapshot after `atomix.stop().join()` is completed.
I'm searching how to implement a minimal reproducer using just `RaftServer` for the PR I've sent to fix this.

 - Atomix: `3.2.0-SNAPSHOT`

franz1981:
Fixes #1096


franz1981:
Hi @kuujo , sorry to bother about this, but we would like to use Atomix as one of the RI for the https://github.com/apache/activemq-artemis distributed lock implementation, but we're hitting this memory leak. 
I would like to bring this into the next Artemis release, so I would be super happy to help providing a test too, not just a fix, let me know what I can do to help. thanks for Atomix is a super nice prj!

johnou:
@franz1981 lgtm, can you add in your test case from the report too please?

franz1981:
@johnou thanks to look into this, sadly the test I've provided won't have any failure to deal with, because what just an example to show how to produce the leak...
I'm now looking how to make it "testable"

johnou:
Silly question but you haven't mixed up shutdown and leave right? shutdown name is a little misleading, it doesn't leave the cluster, while leave should drop from the cluster and shutdown.

franz1981:
AFAIK nope, I've just called a shutdown when the node hasn't yet joined the rest of the cluster (because the other nodes are not reachable, for example, or not yet started too). And was getting the leak after it

franz1981:
Just looked around in the doc and cannot found anything different from `Atomic::stop` for my needs, let me know if there are other methods I've missed....

I see instead, in the doc
````
For example, if the instance is configured with a Raft partition group, a majority of the group’s members must be started for any single member to complete startup.
```
That's my use case: I'm stopping an atomix instance while the majority of group's members are not yet started, because my use case requires to try join the majority and, if failing after N seconds, retry after some time.
FYI the "retry after some time" means creating a new Atomix instance.

The issue I see is that stopping the previous one cause some leaks.

franz1981:
@johnou let me know if I need to share/do anything to help more on this

hbswn:
Has this ever been resolved?
Looks like I ran into the same issue.

Btw, I noticed that the Atomix 3 is gone!

PR #1097, opened 2021-05-10
[AETHER-1265] Integrate Java profiler in TOST env

pierventre:
Fix an issue in the atomix-agent that does prevent passing a different gc algorithm. Adds a new dockerfile that allows to build profiler enabled images.

### DIFF
--- a/dist/src/main/resources/bin/atomix-agent
+++ b/dist/src/main/resources/bin/atomix-agent
@@ -13,8 +13,14 @@ fi
 
 ATOMIX_ROOT="$(dirname "$(dirname "$0")")"
 
+# Pass gc and other useful jvm opts through the env variable
+JAVA_OPTS="${JAVA_OPTS:--XX:+UseG1GC -XX:MaxGCPauseMillis=200}"
+if [ ! -z "$ATOMIX_YOURKIT" ]; then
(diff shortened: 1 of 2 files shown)


dist/src/main/resources/bin/atomix-agent
```diff
@@ -13,8 +13,14 @@ fi
...
 
+# Pass gc and other useful jvm opts through the env variable
+JAVA_OPTS="${JAVA_OPTS:--XX:+UseG1GC -XX:MaxGCPauseMillis=200}"
+if [ ! -z "$ATOMIX_YOURKIT" ]; then
+  JAVA_OPTS+=" -agentpath:$ATOMIX_ROOT/libyjpagent.so=listen=all"
+fi
+export JAVA_OPTS=${JAVA_OPTS}
+
 java \
-  -XX:+UseConcMarkSweepGC \
```
charlesmcchan, line 17: Is this removed intentionally? It seems irrelevant to the profiler change.

pierventre: Done intentionally - otherwise we cannot pass through the JAVA_OPTS a different gc

kuujo (approved):

PR #1101, opened 2021-10-29
Raft take snapshot process is not proper

jbodarya:
Here, the method is taking a snapshot "Per service", but somehow looping over the entire session list. This means at t will create a problem the time of restore, It will create all the sessions for the first service getting loaded.
sample effect :  If there are changes between the take-snapshot and reboot node, All will go missing from the Service.


protocols/raft/src/main/java/io/atomix/protocols/raft/service/RaftServiceContext.java
```diff
@@ -300,8 +300,8 @@ public void takeSnapshot(SnapshotWriter writer) {
     writer.writeLong(currentTimestamp);
     writer.writeLong(timestampDelta);
 
-    writer.writeInt(sessions.getSessions().size());
-    for (RaftSession session : sessions.getSessions()) {
+    writer.writeInt(sessions.getSessions(primitiveId).size());
+    for (RaftSession session : sessions.getSessions(primitiveId)) {
```
johnou, line 304: @jbodarya are you able to write a unit test to demonstrate the problem?

PR #1103, opened 2022-01-20
Add demote API to LeaderElector primitive

pierventre:
The new API allow to demote a non-leader node to the bottom of the candidates list. It is not allowed to displace the current leader. It can be used in conjunction with anoint() to demote the current leader

### DIFF
--- a/core/src/main/java/io/atomix/core/election/impl/DefaultLeaderElectorService.java
+++ b/core/src/main/java/io/atomix/core/election/impl/DefaultLeaderElectorService.java
@@ -188,6 +188,27 @@ public class DefaultLeaderElectorService extends AbstractPrimitiveService<Leader
   }
 
   @Override
+  public boolean demote(String topic, byte[] id) {
+    try {
+      Leadership<byte[]> oldLeadership = leadership(topic);
+      if (oldLeadership == null ||
+              oldLeadership.candidates().stream().noneMatch(candidate -> Arrays.equals(candidate, id)) ||
+              Arrays.equals(oldLeadership.leader().id(), id)) {
+        return false;
+      }
+      elections.computeIfPresent(topic, (k, v) -> v.demote(id));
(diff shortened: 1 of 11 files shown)


core/src/main/java/io/atomix/core/election/impl/DefaultLeaderElectorService.java
```diff
@@ -187,6 +187,27 @@ public boolean promote(String topic, byte[] id) {
...
+  public boolean demote(String topic, byte[] id) {
+    try {
+      Leadership<byte[]> oldLeadership = leadership(topic);
+      if (oldLeadership == null ||
+              oldLeadership.candidates().stream().noneMatch(candidate -> Arrays.equals(candidate, id)) ||
+              Arrays.equals(oldLeadership.leader().id(), id)) {
+        return false;
+      }
+      elections.computeIfPresent(topic, (k, v) -> v.demote(id));
+      Leadership<byte[]> newLeadership = leadership(topic);
```
charlesmcchan, line 200: Is it possible to figure out whether the operation is successful by looking at the `computeIfPresent` return value? Something like:
```
ElectorState newState = elections.computeIfPresent(topic, (k, v) -> v.demote(id));
if (newState != null && !Objects.equal(oldLeadership, newState.leader())) {
  // notify...
}
```

charlesmcchan (approved):

kuujo (approved):

core/src/main/java/io/atomix/core/election/impl/DefaultLeaderElectorService.java
pierventre: I can but I don't have directly the leadership object I want to compare and I still need other code (what `leadership` function does under the hood). I believe it is fine - The service (IIRC) is executed by a single thread we should not worry about race

